PR #3, opened 2023-03-20
Circuit tests

eerkaijun:
Some tests for the `play` and `draw` circuits. Instructions to run the tests are added to the README. New external dependencies needed for the tests `circom-helper`.

### DIFF
--- a/packages/circuits/circom.config.json
+++ b/packages/circuits/circom.config.json
@@ -1,0 +1,7 @@
+{
+    "circom": "../../../../.cargo/bin/circom",
+    "snarkjs": "./node_modules/snarkjs/build/cli.cjs",
+    "circuitDirs": [
+        "./src"
+    ]
+}
--- a/packages/circuits/circom.config.json
+++ b/packages/circuits/circom.config.json
@@ -2,6 +2,6 @@
     "circom": "../../../../.cargo/bin/circom",
     "snarkjs": "./node_modules/snarkjs/build/cli.cjs",
     "circuitDirs": [
-        "./src"
+        "./test/circuits"
     ]
 }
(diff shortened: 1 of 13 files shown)


packages/circuits/src/merkle.circom
```diff
@@ -13,7 +13,7 @@ template HashLeftRight() {
     signal output hash;
 
     // TODO: use more rounds
-    component hasher = MiMCSponge(2, 4, 1);
+    component hasher = MiMCSponge(2, 220, 1);
```
norswap, line 16: Not really a concern for this, but did you have a look at proof generation times?

packages/circuits/circom.config.json
```diff
@@ -0,0 +1,7 @@
+{
+    "circom": "../../../../.cargo/bin/circom",
```
norswap, line 2: I wonder if we could easily populate this from `which circom`

`circom-helper` doesn't seem to have an option to set this directly, but we could replace this with a script that creates the config file on demand.

packages/circuits/package.json
```diff
@@ -4,7 +4,17 @@
...
   "dependencies": {
+    "circom-helper": "^0.3.5",
     "circomlib": "^2.0.5",
+    "circomlibjs": "^0.1.7",
+    "ffjavascript": "^0.2.57",
     "snarkjs": "^0.6.9"
+  },
+  "scripts": {
+    "circom-helper": "npx circom-helper -c ./circom.config.json -b ./build/test -p 9001",
+    "test": "jest"
```
norswap, line 15: I try to consolidate all useful commands in Makefiles, but I'll take care of this!

PR #8, opened 2023-03-26
Initial hand proof

eerkaijun:
Added a new circuit that proves that a user has correctly drawn cards from the deck initially. Included a test to check that the circuit works. Also generated a smart contract that can verify this proof on-chain.


packages/circuits/src/initial.circom
```diff
@@ -0,0 +1,89 @@
...
+    signal output newArray[n];
+    signal output replacedValue;
+
+    component checkIndex[n];
+    signal items[n];
+    for (var i = 0; i < n; i++) {
+        checkIndex[i] = IsEqual();
+        checkIndex[i].in[0] <== i;
+        checkIndex[i].in[1] <== index;
+        newArray[i] <== (newValue - array[i]) * checkIndex[i].out + array[i];
```
norswap, line 19: Maybe comment the line above with `newArray[i] = (i == index) ? newValue : array[i]`?

packages/circuits/src/initial.circom
```diff
@@ -0,0 +1,89 @@
...
+        // newArray[i] <== (checkIndex[i].out * newValue) + ((1 - checkIndex[i].out) * array[i]);
+        items[i] <== checkIndex[i].out * array[i];
+    }
+
+    // get the replaced value
+    var temp = 0;
+    for (var i = 0; i < n; i++) {
+        temp += items[i];
+    }
+    replacedValue <== temp;
```
norswap, line 29: Is this safe? My understanding of circom is that you could set temp to anything you want here (i.e. it is not bound to `array[index]`).

Can't we simply include `replacedValue <== array[index]` here?

packages/circuits/src/initial.circom
```diff
@@ -0,0 +1,89 @@
...
+    checkInitialDeck.leaves <== deckLeaves;
+
+    // check initial hand is all null
+    var handLeaves[2**levels];
+    for (var i = 0; i < 2**levels; i++) {
+        handLeaves[i] = 255;
+    }
+    component checkInitialHand = CheckMerkleRoot(levels);
+    checkInitialHand.root <== handRoot;
+    checkInitialHand.leaves <== handLeaves;
```
norswap, line 52: I don't think we need this check (that the initial hand is all null) — the hand will be initialized when the game is joined, and the hand is never written before. I will in fact *not* be equal to a root of all nulls, but instead to 0.

We can simply skip the `handRoot` signal altogether.

packages/circuits/test/circuits/initial.test.circom
```diff
@@ -0,0 +1,5 @@
+pragma circom 2.0.0;
+
+include "../../src/initial.circom";
+
+component main {public [deckRoot, newDeckRoot, initialDeckTailCardIndex, handRoot, newHandRoot, drawnCardIndices]} = Initial(4, 2);
```
norswap, line 5: Any insight on the performance of this? What about bigger numbers? (I'd be particularly interested in 6 levels (64 cards) and a hand of 7 cards or so which mimics the deck/hand sizes in popular card games.)

packages/circuits/test/draw.test.js
```diff
@@ -35,6 +35,8 @@ describe("Draw Cards Test", () => {
         handRoot = mimcsponge.multiHash([handHash1, handHash2]);
     });
 
+    // set longer timeout for test
+    jest.setTimeout(10000);
```
norswap, line 39: What does this do exactly? (Note I'm not familiar with Jest)

packages/circuits/src/merkle.circom
```diff
@@ -191,4 +191,27 @@ template ConstructRoot(levels) {
     }
 
     root <== hashers[levels - 1].hash;
+}
+
+template CheckMerkleRoot(levels) {
+    signal input root;
+    signal input leaves[2**levels];
```
norswap, line 198: Big nitpick, but when I imagine a "two-level Merkle tree", I imagine one with 3 nodes (first level = root, second level = leaves). This seems to take `levels` as being the non-root levels?

packages/circuits/src/initial.circom
```diff
@@ -0,0 +1,89 @@
...
+}
+
+template Initial(levels, cardCount) {
+    signal input deckRoot;
+    signal input newDeckRoot;
+    signal input deckLeaves[2**levels];
+    signal input initialDeckTailCardIndex;
+    signal input handRoot;
+    signal input newHandRoot;
+    signal input drawnCardIndices[cardCount];
```
norswap, line 39: This is missing the randomness as an input, and in general the verification that the indices were selected correctly.

However, this can probably be added in another component that then includes this component!

packages/circuits/src/initial.circom
norswap: Actually I see below that you use `something <== x ? y : z`, can't we use the same pattern here?

I'm thinking maybe not, because below, `x` is only dependent on the loop index, and not on the signal values.

packages/circuits/src/initial.circom
norswap: Ah! Maybe it's secure because `var`is under the hood an array of signals that get constrained to their next value. In any case, the question about the simpler form still holds.

packages/circuits/src/initial.circom
```diff
@@ -0,0 +1,89 @@
...
+        updateDeckArray[i+1].array <== updateDeckArray[i].newArray;
+
+        // update tail card index
+        tailCardIndex--;
+
+        // update hand with the drawn card
+        updateHandArray[i/2] = CheckReplaceArray(2**levels);
+        updateHandArray[i/2].index <== i/2;
+        updateHandArray[i/2].newValue <== updateDeckArray[i+1].replacedValue;
+        updateHandArray[i/2].array <== i == 0 ? handLeaves : updateHandArray[i/2-1].newArray;
```
norswap, line 77: Couldn't we get rid of all the iterations over the deck & hand? Currently this iterates over all drawn cards, and then for every one of them generates two iterations over the whole deck and one over the hand (via three `CheckReplaceArray` components) to get the updated deck & hand.

What if instead we still iterate over the cards drawn, but directly set the values in an `updatedDeck` and an `updatedHand` array variables?

packages/circuits/src/initial.circom
```diff
@@ -0,0 +1,89 @@
...
+    signal output replacedValue;
+
+    component checkIndex[n];
+    signal items[n];
+    for (var i = 0; i < n; i++) {
+        checkIndex[i] = IsEqual();
+        checkIndex[i].in[0] <== i;
+        checkIndex[i].in[1] <== index;
+        newArray[i] <== (newValue - array[i]) * checkIndex[i].out + array[i];
+        // newArray[i] <== (checkIndex[i].out * newValue) + ((1 - checkIndex[i].out) * array[i]);
```
norswap, line 20: Get rid of comment :)

packages/circuits/test/circuits/initial.test.circom
eerkaijun: Good point, let me try running some different values and compare the performance.

packages/circuits/test/draw.test.js
eerkaijun: Jest by default has a 5000ms timeout, so if the test does not complete within 5s, it's considered as fail. Since some of this proof generation takes a little longer, I set it to a longer period to make sure they can be run to complete.

packages/circuits/src/initial.circom
eerkaijun: Yeap, that's right - basically any conditional statement in Circom cannot uses an unknown value during compilation, otherwise it will run into non quadratic constraint error.

packages/circuits/src/initial.circom
eerkaijun: We can't use `replacedValue <== array[index]` because `index` is an unknown value during compilation, where in Circom any reading/writing to an array couldn't use an unknown value, otherwise there will be non-quadratic constraint error. I do feel that my approach of looping through the array and summing to `temp` is pretty dumb, but I couldn't think of a better way, so if there's other ideas or inspirations that would be appreciated too.

packages/circuits/src/initial.circom
eerkaijun: That makes sense, will make the change!

packages/circuits/src/merkle.circom
eerkaijun: That's what I thought initially too, but seems like the other circuits use `levels` as non-root levels. Thinking a bit deeper, I think it might make sense so we can define all the adjacent hash path arrays as size `levels` rather than size `levels - 1`. What do you think?

packages/circuits/src/initial.circom
eerkaijun: That was my first approach, but apparently an annoying thing about Circom is that all arrays cannot use values that are unknown during compilation, which will lead to non quadratic constraint error. So for example if you ever assign something like `updateDeck[index] = x`, it will become non quadratic. 

I do feel that my approach is actually pretty inefficient hahaha, so keen to get maybe @Boffee thoughts on this. (Kudos to him for pointing out about the array index issue, I was stuck on that for a long time TT)

packages/circuits/src/initial.circom
norswap: Right, that seems like a pretty fundamental limitation. Circomlib has [`Mux4`](https://github.com/iden3/circomlib/blob/master/circuits/mux4.circom) that allows you to index into an array of size 16, but that wouldn't work for the deck.

One question is whether we need `replacedValue` at all. You use it here: 

```
updateDeckArray[i+1].newValue <== updateDeckArray[i].replacedValue;
```

But I think we could skip this if we supplied the list of drawn cards as a signal. We'd need an additional constraint to check that the card matches the index, but we'd avoid O(deck_size) extra constraints for getting the replaced value.

packages/circuits/src/initial.circom
norswap: I spent some time thinking on it, I feel like a better approach here might be to iterate the draw proof as many times as there are cards to draw.

Each draw proof basically needs to do a Merkle proof for the card to remove + for the last card (`tailCard` in the implem), as well as construct two Merkle roots (one for copying the tail card to the removed card index, one to remove the tail card). Each of these operation have log complexity in the size of the deck, so you pay `O(4 * initialHandSize * log(deckSize))` where we can constrain `log(deckSize)`to 6 in practice (assuming max 64 cards per deck). So for a hand of 7 cards and a deck of 60 cards, `168c` where `c` is some constant number of ops from the implem of membership checks and root construction.

In contrast, the current approach, does `O(2 * deckSize * initialHandSize)` so `840c` (of course one also need to check to see if the overhead is different ...).

A naive approach would need Merkle branch for every removed card + tailCard to be provided, but in reality we can recompute the whole Merkle tree ourselves + compute the Merkle branch dynamically (zero constraints for Merkle branch computation, using `<--` because everything is ultimately authenticated against the Merkle roots, and each successive Merkle root is correct by construction).

By the way, I looked at the code for the draw proof, and there are quite some problems there also — for one, I found a pretty bad bug I think (the code does not actually computes the second Merkle root as it should, instead I just does a membership check against it). It also does not ensure that the `tailCard` that is passed is actually the last card in the deck (this could be fine if we say that the remaining deck size is tracked in the contract and use that to supply the last card index to the verifier).

norswap: As is often the case, I had to post this to realize how stupid it was: the overhead from the hashing in the Merkle root probably makes the 4-5x gain obsolete. Oh well.

packages/circuits/src/initial.circom
eerkaijun: If that's the case, we will need to pass in both the drawn cards and tail cards as signals, as we need both to check that the decks have been updated correctly. Could again do a performance comparison.

packages/circuits/src/initial.circom
eerkaijun: Just did a comparison for using Merkle root as temporary state instead of using a deck array. The constraints of the circuits are as follows: 
1. Deck array update (current approach): ~180k constraints
2. Merkle root update : ~390k constraints

From what I understand, number of constraints is an indication of optimization of circuit, so perhaps the merkle approach is more expensive. 180k is still a large circuit though, for comparison our current other circuits draw.circom is at ~55k and play.circom is at ~31k

packages/circuits/test/circuits/initial.test.circom
eerkaijun: Okay tried it out! Proof generation takes around 22-25 seconds averagely for 6 levels and 7 cards. More detailed constraint numbers are in comment below.

norswap: Nice work! But damn, that's way too much for the game :D

I'm guessing the shuffling idea Brian suggested is the way to go: use a proof of shuffle, then we can have the contract compute all random (unique!) draw indices and pass them into the initial hand proof and use the bitfield approach to reduce to a single pass over the deck and a single pass over the hand.

This will require changing the contracts so that there the randomness seed for the initial hand is only known after the shuffle is committed to.

eerkaijun:
@norswap let me know if you have any feedback on the latest circuit!

packages/circuits/src/merkle.circom
norswap, line 1: I wonder to what extent we couldn't replace a lot of what is in this file with the implementation from circomlib: 
https://github.com/iden3/circomlib/blob/master/circuits/smt/smtprocessor.circom

(Crucially, we need to verify how it impacts proving time, given equivalent hash functions.)

The thing linked above handles checking for Merkle tree insertion/update/deletion.
Might require us to change how we represent empty leaves however!

There does not seem to be a Merkle root creator from a list of leafs, but it's hard to see how to do that more efficiently than how it's done here however.

packages/circuits/src/initial.circom
```diff
@@ -0,0 +1,61 @@
...
+
+template Initial(levels, cardCount) {
+    signal input deckRoot;
+    signal input newDeckRoot;
+    signal input deckLeaves[2**levels];
+    signal input newDeckLeaves[2**levels];
+    signal input newHandRoot;
+    signal input newHandLeaves[2**levels];
+    signal input deckPredicate[2**levels];
+    signal input drawnCardIndices[2**levels];
```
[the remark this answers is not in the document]
norswap: This could use some documentation:
- `levels` is the base-2 log of the max size of the deck (6 for a deck of max 64 cards)
- `cardCount` is the amount of cards to draw
- `deckRoot` is the root of the initial deck listing
- `deckLeaves` are the cards in the initial deck listing (padded with 255 to reach a size of `2**levels`)
- `newDeckRoot` is the root of the deck after drawing the cards and unconstrained shuffling by the player
- `newDeckLeaves` are the cards in the new deck (cf. `newDeckRoot`), padded with 255 to reach a size of `2**levels`
- `drawnCardIndices` is an array contain containing 1 if the corresponding index is to be drawn, zero otherwise.
- `drawnPredicate[i]` contains
     - if `drawnCardIndices[i] == 0` : `(2**levels) ** positionInHand`
     - if `drawnCardIndices[i] == 1` : `(2**levels) ** positionInNewDeck`
     (positions start at 0)
     
This maintains privacy because the users shuffles his deck after drawing the cards and does not reveal this new ordering.

packages/circuits/src/initial.circom
```diff
@@ -0,0 +1,61 @@
...
+        mux[i].in[0] <== deckLeaves[i] * deckPredicate[i];
+        mux[i].in[1] <== 0;
+        mux[i].s <== drawnCardIndices[i];
+        deckSum += mux[i].out[0];
+        handSum += mux[i].out[1];
+    }
+    // check that number of drawn cards is correct
+    cardCount === indicesSum;
+
+    // check that the cards in the new deck is correct
```
norswap, line 40:
```suggestion
    // check that the new deck contains the same cards as the old deck, minus those that were drawn
```

packages/circuits/src/initial.circom
```diff
@@ -0,0 +1,61 @@
...
+    var newDeckSum = 0;
+    for (var i = 0; i < 2**levels - cardCount; i++) {
+        newDeckSum += newDeckLeaves[i] * (maxDeckSize**i);
+    }
+    deckSum === newDeckSum;
+    // check that the remaining deck is null
```
norswap, line 46:
```suggestion
    // check that the remaining of the deck is null
```

packages/circuits/test/initial.test.js
```diff
@@ -0,0 +1,101 @@
...
+            if (drawnIndices.has(i)) {
+                // card is drawn
+                deckPredicate.push(maxDeckSize ** handCount);
+                handCount++;
+            } else {
+                // card is not drawn
+                deckPredicate.push(maxDeckSize ** deckCount);
+                deckCount++;
+            }
+        }
```
norswap, line 55: It would be good to test some shuffling here.

packages/circuits/src/merkle.circom
norswap: I'm also intrigued that they are using  `MultiMiMC7` instead of `MiMCSponge` ... I wonder if that gives better performance?

However, it seems like they add an extra layer of hashing by hashing leaves as `H(a | key (index) | value)`, so maybe not so good...

packages/circuits/src/initial.circom
```diff
@@ -0,0 +1,61 @@
...
+    signal input newHandRoot;
+    signal input newHandLeaves[2**levels];
+    signal input deckPredicate[2**levels];
+    signal input drawnCardIndices[2**levels];
+
+    var maxDeckSize = 2**levels;
+
+    component checkInitialDeck = CheckMerkleRoot(levels);
+    checkInitialDeck.root <== deckRoot;
+    checkInitialDeck.leaves <== deckLeaves;
```
norswap, line 19: We need to check the new deck root in the same way, no?

Also now that I think on it... This must already be a shuffled deck, right? Otherwise everyone knows what will be drawn.

This means we need another proof, that is a pure proof of shuffle wrt a known Merkle proof?
This is easy to implement (essentially, it's just a subset of this proof!).

This could potentially be done ahead of the game though, and could even be reused for multiple game.

I'm actually starting to think I'd love closed deck lists, but then a nasty side-effect is that the proof for every deck needs to be regenerated when cards are being traded (or moved out of the inventory).

norswap (commented):
Daaamn, this is elegant!

packages/circuits/src/merkle.circom
eerkaijun: Makes sense :)

packages/circuits/src/merkle.circom
eerkaijun: Yes, `MultiMiMC7` is faster than `MiMCSponge` because it has less rounds. I am wondering how do you guys decide which hash function to use? Why not others like `Poseidon`?

norswap: @Boffee will be able to tell you more, but IIRC Mimc was faster than Posseidon?

norswap: I'm guessing the difference is not as simple as a number of rounds? (If so, why not use the MimcSponge template w/ a lower round number?)

packages/circuits/test/initial.test.js
```diff
@@ -21,8 +21,8 @@ describe("Initial Hand Test", () => {
     });
 
     // set longer timeout for test
-    jest.setTimeout(10000);
-    it("Should correctly construct a merkle tree root", async () => {
+    jest.setTimeout(25000);
```
eerkaijun, line 24: Increasing the timeout because running more tests in parallel make each test slower too. Not sure if jest can run tests sequentially...

packages/circuits/test/initial.test.js
eerkaijun: Added the shuffle test in another file :)

packages/circuits/src/initial.circom
eerkaijun: Yeap have now added a shuffle circuit. The workflow I am envisioning is:
1. opened deck list added
2. player shuffle the deck to hide the order of the cards, and committing a merkle root for the new order
3. random drawn indices generated by the contract, which is public, and the player draws the card based on these indices
4. player commits new deck root and new hand root

Just wanted to make sure I am thinking correctly?

packages/circuits/src/shuffle.circom
```diff
@@ -0,0 +1,35 @@
+pragma circom 2.0.0;
+
+include "./merkle.circom";
+
+template Shuffle(levels) {
+
+    signal input initialDeck[2**levels];
+    signal input finalDeck[2**levels];
+    signal input deckPredicate[2**levels];
```
norswap, line 9: Would be good to define (comment) `deckPredicate` here.

I'm also not in love with the name (also in the other proof). Maybe `initialDeckWeights`? Open to suggestions!

packages/circuits/src/shuffle.circom
```diff
@@ -0,0 +1,35 @@
...
+    for (var i = 0; i < maxDeckSize; i++) {
+        deckSum += initialDeck[i] * (maxDeckSize**i);
+    }
+
+    signal tempSum[2**levels]; // we need to use intermediate signal to prevent non quadratic error
+    for (var i = 0; i < maxDeckSize; i++) {
+        if (i == 0) {
+            tempSum[i] <== finalDeck[i];
+        } else {
+            tempSum[i] <== tempSum[i-1] + (finalDeck[i] * deckPredicate[i]);
```
norswap, line 24: So I'm thinking ... is this even secure? 😅 

To prove the shuffle `[x, y, z] → [x, y, z]`, you would prove

```x * 1 + y * 64 + z * 64^2 ==
x * 1 + z * 64 ^2 + y * 64
```

But then, you could prove the (incorrect) shuffle `[x, y, z] → [x, y, w]` by proving
```
x * 1 + y * 64 + z * 64^2 ==
x * 1 + y * 64 + w * (z * 64^2 / w)
```

I think the solution is simply to constrain the `deckPredicate` array by computing `sum(deckSize ** i)` and ensuring the sum over `deckPredicate` is equal to that.

Googled a bit and found this: https://zkresear.ch/t/random-permutation-circuit/74
It uses a random value (`hash`) to shuffle the deck. We could allow just any value to be passed for that. Might be a lot more efficient than what we have right now (4.3k constraints says Jordi).

Haven't studied the circuit yet but I intend to! Wonder if that could also be used to reduce the complexity of the original hand proof.

norswap (commented):
Have to catch my plane, but posting this so info is already available.

packages/circuits/test/initial.test.js
norswap: It does test another circuit though (which is not reused by the `Initial` circuit) so it would be easily possible for the circuit not to check the shuffling (the predicates) correctly :p

packages/circuits/src/initial.circom
eerkaijun: Oh yeah good catch hahaha, fixed the `newDeckRoot` issue.

packages/circuits/test/initial.test.js
eerkaijun: That makes sense, have made the changes.

packages/circuits/src/shuffle.circom
eerkaijun: I think `deckWeights` might make sense, since it's not really "initial" per se as the weights are dependent on the drawn indices.

packages/circuits/src/shuffle.circom
eerkaijun: Oooh this is a very interesting observation! I have constrained the deck predicate for now - thanks for sharing the permutation circuit, checking it out..

norswap:
This is pretty good for now! Let's merge, I created the following issues for followup:
- https://github.com/norswap/0xFable/issues/10
- https://github.com/norswap/0xFable/issues/11

PR #9, opened 2023-03-29
Collection retrieved from contracts, cards displayed in grid, card selection, search & filtering, empty collection page

geniusgarlic:
Stuff I'm not sure about: 
-the handling of the override in 'InventoryCardsCollection.sol'
-the way I get the lore from 'CardsCollection.sol' feels bad
-maybe the way I'm getting the collection from the contracts in the frontend isn't the cleanest either


packages/webapp/src/components/modals/collectionModal.tsx
norswap, line 1: This does not need to be its own component. The other buttons are because they include a modal (this means "pop-up"!) with some logic, but this is literally just a button, so better to inline it.

packages/webapp/src/components/modals/collectionModal.tsx
```diff
@@ -0,0 +1,24 @@
+import { useRouter } from 'next/router';
+
+export const CollectionGameModal = () => {
+
+  const router = useRouter();
+  const handleClick = () => {
+    router.push('/collection');
+  };
```
norswap, line 8: You can inline this: `<label ... onClick={() => router.push('/collection')}`.

In fact it might be better to use a Next.js `Link` instead of `label`. More on this below.

packages/webapp/src/pages/collection.tsx
```diff
@@ -1,8 +1,62 @@
 import { type NextPage } from "next";
 import Head from "next/head";
 import { Navbar } from "../components/navbar";
+import React, { useState, useEffect } from 'react';
+import { useInventoryGetCollection } from "../generated";
```
norswap, line 5: We haven't done this everywhere, but we should start:

1. Making all imports from files in the project start with `src` instead of being relative. (Sasha made this possible with the typescript config. So `src/generated.tsx` instead of `../generated`. One advantage is the import will stay correct if we move the file around.
2. Let's separate the imports from project files and the imports from files from libraries (e.g. start with all the library imports, then all the local file imports)
3. (less important) avoid semicolons — I will probably look into linters and enforce this automatically at some point

packages/webapp/src/styles/globals.css
norswap, line 1: Let's try to avoid adding CSS (unless where it's really cleaner / saves a lot of repeated text) and use tailwind CSS properties to style components. Here is the tailwind CSS docs: https://tailwindcss.com/docs

In particular, all the flexbox stuff should be easily doable within tailwind.

We also use DaisyUI, which is a theming / components library for Tailwind. We use the "coffee" theme.
These themes come with default colors that we could try using, but also properties that can be used to specify a bunch of colors: https://daisyui.com/docs/colors/

norswap (commented):
Partial review (have to run) — I still need to review `collection.tsx` + checkout locally to see what this looks like.

Good job so far, looks like a meaningful chunk of work! :)

norswap:
Screenshot

Screenshot 2023-04-03 at 14 29 30

packages/webapp/src/pages/collection.tsx
```diff
@@ -14,8 +68,96 @@ const Play: NextPage = () => {
...
+              <div className="h-20"></div> {/*to add padding at the bottom*/}
+            </div>
+
+        </div>
+
+          {/* the actual card collection displayed on the right*/}
+          <div className="col-span-9 h-full rounded-xl border overflow-auto h-screen">
+            { !cardsAvailable &&
+              <div className="flex flex-col justify-center items-center h-screen overscroll-auto">
+                {/*Can't remove the scrollba*/}
```
norswap, line 125: What do you mean?

packages/webapp/src/pages/collection.tsx
```diff
@@ -1,8 +1,62 @@
 import { type NextPage } from "next";
 import Head from "next/head";
 import { Navbar } from "../components/navbar";
+import React, { useState, useEffect } from 'react';
+import { useInventoryGetCollection } from "../generated";
+import { useAccount } from 'wagmi'
+import { deployment } from "src/deployment";
+// Eventually, you have to get all the effects used in the collection's cards
+const effects = [{label: 'Charge', active: false}, {label: 'Flight', active: false}, {label: 'Courage', active: false}, {label: 'Undying', active: false}, {label: 'Frenzy', active: false}, {label: 'Enlightened', active: false}];
+const types = [{label: 'Creature', active: false}, {label: 'Magic', active: false}, {label: 'Weapon', active: false}];
```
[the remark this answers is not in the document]
geniusgarlic: Everything is visible since the filtering is not activated by default. The main problem is that the cards that are minted are mainly identical: they have the same name and don't have effects and types yet. I'll try to deal with that after fixing the rest of the collection.

packages/webapp/src/pages/collection.tsx
```diff
@@ -1,8 +1,74 @@
...
+    const newEffects = [...activeEffects];
+    newEffects[index].active = !newEffects[index].active;
+    setActiveEffects(newEffects);
+  };
+
+  const handleTypeClick = (index) => {
+    console.log(cards);
+    console.log(typeof cards === 'undefined');
+    console.log(cardsAvailable);
+    console.log(!!cards && cards.length > 0);
```
norswap, line 47: Remove logs

packages/webapp/src/pages/collection.tsx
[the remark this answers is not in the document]
norswap: Hhm I see, do we can either say: by default all the filters are on `true`, or by default, filtering is off and we when we switch it on we only want to display one thing. I've definitely seen both on websites. I guess the way you implemented it makes more sense here.

packages/contracts/src/Inventory.sol
```diff
@@ -247,4 +247,16 @@ contract Inventory {
...
+    // Return the list of cards in the collection of the given player.
+    function getCollection(address player) external view returns(Card[] memory collectionCards) {
+        uint256[] memory collectionTokensId = inventoryCardsCollection.getOwnedTokens(player);
+        collectionCards = new Card[](collectionTokensId.length);
+        for (uint256 i = 0; i < collectionTokensId.length; ++i) {
+            collectionCards[i].lore = originalCardsCollection.getLore(collectionTokensId[i]);
+            collectionCards[i].stats = originalCardsCollection.stats(collectionTokensId[i]);
+        }
+        return collectionCards;
+    }
```
norswap, line 261: - [] Should move to `InventoryCardsCollection`

norswap:
Great job! Merging this :)

PR #12, opened 2023-04-15
massive frontend refactor & more

norswap:


packages/webapp/src/pages/collection.tsx
norswap, line 1: @geniusgarlic You may want to have a look at the changes here!

PR #14, opened 2023-05-25
Initial hand verification

eerkaijun:
Swapped out groth16 verifier to plonk verifier. 

Actually I noticed an issue for the initial hand proof, which is we calculate the `deckPredicate` as `deckSize ** deckCount` (https://github.com/norswap/0xFable/blob/master/packages/circuits/src/initial.circom#L69), and some of the values will actually be larger than what can be stored in `uint256`, so it won't work on the smart contract part. Perhaps we could use modulo for the `deckPredicate`?


norswap:
What caused it to be compiled with Groth instead of Plonk? Manual commands vs Makefile? Testing setup? Is there anything we can change/add in the Makefile to avoid this kind of problem?

Good catch regarding the predicate. Seems like the smallest `x` such that `x^x / 2^256 < 1` is 46. That might be too small for comfort.

However, it seems that arithmetic is already done modulus a large prime: https://docs.circom.io/circom-language/basic-operators/ so we should be good there.

That being said, looking at the code, it looks like `deckPredicateSum` (unlike `tempDeckPredicateSum`) is unconstrained, meaning it would be possible to generate a proof for any set of `deckPredicates` passed in. Am I missing something?

If that's true, it also applies to the various sums, the only constraint being on e.g. `deckSum === newDeckSum`. But the construction of `deckSum` and `newDeckSum` being unconstrained makes proving this trivial with any value.

If that's true also, it also means the circuits takes a lot more time to prove than we thought :')

norswap:
Oh and I noticed that [this](https://github.com/norswap/0xFable/blob/master/packages/circuits/src/initial.circom#L14) should say `deckPredicate`

eerkaijun:
I think I was previously manually generating the verification smart contract, that's why end up being in groth 16. 

For the `deckPredicateSum`, the value is summed up in the circuit itself, and it is checking that (e.g. in the example of 64 cards and 7 draws):
- sum of (`maxDeckSize ** i`), where i = 0 to 56 inclusive
- sum of (`maxDeckSize ** i`), where i = 0 to 6 inclusive

So I think that is already the constraint since it's constant? And the `deckPredicateSum` is subsequently used to constraint the `tempDeckPredicateSum`.

norswap:
So my understanding (which I think is correct) is that your Circom program basically defines two things:
- a set of constraints (only `===`, `==>` and `<==`), and a prover
- the prover is used to generate witnesses that satisfy the constraint

But the code that the prover runs is not itself constrained, you can bypass the prover entirely and inject your own witness values. Meaning that you can't rely on the computation done by the prover for correctness.

One way to thing about this: constraints are saying "there exists x and y, such that `x === y * 2 + 3` (etc etc)" while the prover gives an algorithm to derive these existing values.

An example: if you could rely on the prover generation code, proving hashes would be trivial: you'd compute the hash from the preimage in the prover then have a single constraint equating it to the verified hash.

eerkaijun:
Interesting - in that case, I think we can pass `deckPredicateSum` as a public input into the circuit, and we constraint that in the smart contract?

norswap:
I don't even need you need to pass it as public input, as long as you constraint its computation.

So like there should be a `signal maxDeckSizeExponents[maxDeckSize - cardCount]` where `maxDeckSizeExponent[i] == maxDeckSize^i` computed as `maxDeckSize[i-1]*maxDeckSize`, and a `signal deckPredicateRunningSum[maxDeckSize - cardCount]` where `deckPredicateRunningSum[i] = deckPredicateRunningSum[i - 1] + maxDeckSizeExponent[i]`.

And probably the same for `handSum` and `deckSum`. I want to reimmerse myself in that code a little bit, each time I look I sort of forgot what all these things are checking and how they relate to each other.

eerkaijun:
Aaa okay I get what you mean...

packages/circuits/src/initial.circom
```diff
@@ -65,8 +63,8 @@ template Initial(levels, cardCount) {
     var deckPredicateSum = 0;
     var newDeckSum = 0;
     for (var i = 0; i < 2**levels - cardCount; i++) {
-        newDeckSum += newDeckLeaves[i] * (maxDeckSize**i);
-        deckPredicateSum += (maxDeckSize**i);
+        newDeckSum += newDeckLeaves[i] * (2**i);
+        deckPredicateSum += (2**i);
```
norswap, line 67: Is this correct? I think there's a problem where card with ID 2 at position 8 and card with ID 8 at position 2 would end up counting for the same weight in the sum, hence not constraining the deck as we want, right?

packages/contracts/src/Game.sol
```diff
@@ -467,9 +467,59 @@ contract Game {
...
+                pubSignals[i] = 2**handCount;
+                handCount++;
+            } else {
+                // the index is not drawn
+                pubSignals[i] = 2**deckCount;
+                deckCount++;
+            }
+        }
+        initialVerifier.verifyProof(proof, pubSignals);
+    }
```
norswap, line 499: Nice! We probably want to parameterize with a max deck size constant but this is more than good enough for now.

packages/contracts/src/Game.sol
```diff
@@ -467,9 +467,59 @@ contract Game {
...
+            // Generate a random index within the remaining elements
+            uint256 randomIndex = uint256(keccak256(abi.encodePacked(randomness, i))) % (lastIndex + 1);
+            
+            // Select the random element and move the last element to its place
+            selectedIndices[i] = indices[randomIndex];
+            indices[randomIndex] = indices[lastIndex];
+            lastIndex--;
+        }
+        
+        return selectedIndices;
```
norswap, line 522: This cannot actually live in the contract and must be implemented in the circuit side. The reason is that knowledge of the drawn indices means the opponent would know the deck.

To make this secure, we actually need the user to also pick a salt, and submit the hash of the salt in a first tx. Then later he can submit the proof, where the randomness (computed on the circuit side) will be based on both the salt that has been committed to and the randomness picked from the contract by using the hash for the last block number.

(It's crucial that the user does not know which hash will be used when he picks his salt, otherwise he could grind the salt to get the hand he wants.)

The circuit will take the salt as private input, its hash as public input and verify adequation. It also take "the public randomness" as public input, and computes the cards to draw based on this.

norswap:
> I don't even need you need to pass it as public input, as long as you constraint its computation.
> 
> So like there should be a `signal maxDeckSizeExponents[maxDeckSize - cardCount]` where `maxDeckSizeExponent[i] == maxDeckSize^i` computed as `maxDeckSize[i-1]*maxDeckSize`, and a `signal deckPredicateRunningSum[maxDeckSize - cardCount]` where `deckPredicateRunningSum[i] = deckPredicateRunningSum[i - 1] + maxDeckSizeExponent[i]`.
> 
> And probably the same for `handSum` and `deckSum`. I want to reimmerse myself in that code a little bit, each time I look I sort of forgot what all these things are checking and how they relate to each other.

To provide some closure on this, Kai pointed out that the [Circom doc](https://docs.circom.io/circom-language/constraint-generation/) seems to imply that any number of additive terms can be used in linear expressions, and as per the example at the end of the page, `+=` seems to be a perfectly valid way to build these linear expressions.

So in principle, we're good and I was worrying for nothing.

packages/circuits/src/initial.circom
eerkaijun: Oh yeah, there could be collision, didn't think of that. The reason that I change it to 2 instead of using `maxDeckSize` is because of overflow issue on the smart contract side (i.e. when we do calculation of `64**64` to get the `deckPredicate`). Even if we use modulo, we still cannot run away from needing to do these calculations. What do you think could be a good way to tackle this?

norswap: Why can't we use the modulus on the smart contract too?

But also ... do we need to make these calculations on the contract side? I think this can be private signals if we pick caids in the circuits (which we must do, cf. other comment)

packages/contracts/src/Game.sol
eerkaijun: I see, my initial thought is that when a player has its deck at the start of the game, it will first shuffle the deck to hide the ordering from the opponent (there's a `shuffle` circuit to make sure the player has shuffle correctly, the player could shuffle arbitrarily). Then, we will draw the cards randomly and publicly on the smart contract. 

I think what you mentioned also works, but implementing the fisher yates algorithm to select cards in the circuit could be expensive, since it involves accessing an unknown index during compilation. This usually requires looping through the entire array for each selection. So I was thinking it might be easier to shuffle off-chain, verify it through the `shuffle` circuit and draw the indices publicly. But I guess this is dependent on the desired game flow, so could also make the changes. Let me know your thoughts!

packages/circuits/src/initial.circom
eerkaijun: For example, if on this [line](https://github.com/norswap/0xFable/blob/initial-hand-verification/packages/contracts/src/Game.sol#L494), using a modulus would be like `pubSignals[i] = 64**deckCount % (2**256)` if I'm thinking correctly? Needing to calculate `64**deckCount` already overflows although the final value is not.

packages/contracts/src/Game.sol
norswap: Oh lord, I don't remember the `Shuffle` circuit at all :')

You don't need to shuffle at all though! If your indices are private (which they are if you use the salt in the randomness) then it doesn't matter that the deck is not initially shuffled — there is not way for your opponent to know what you drew.

(Now that I think of it, he might be able to brute force the Merkle root, so it should probably be hashed with the salt before being posted on chain too.)

packages/circuits/src/initial.circom
norswap: You would need to use the exact prime that Circom uses (and not 2^256) otherwise the contract vs circuit calculation wouldn't match (but again, I don't think this needs to be computed on the contract at all).

In general, you want to use a large prime to compute these checksums, as using something that is not can lead to collisions.

packages/contracts/src/Game.sol
eerkaijun: But to prevent for selecting the same indices twice, we still need to use the fisher yates algorithm to move the selected card to the last index? Essentially I think this [function](https://github.com/norswap/0xFable/blob/initial-hand-verification/packages/contracts/src/Game.sol#L501) is a form of fisher yates, and if we move this function to the circuit, we'll need to: 
- introduce `signal` of size `7 x 64`, where each row is used for each card selection
- for each selection, loop through the entire deck of 64 cards

norswap: Yes, that seems to be correct. There might be a better solution, but it there is I can't think of it right now!

There might be hope if we can do the "bitfield trick" where we represent an array of size `S <= 256` of items of magnitude `M < 256` as a bitfield and can use that bitfield to test if an item is in the array in a constant number of constraints (vs iterating over the whole array, which we would have had to do with no bitfield). Did we ever try if that works?

(Still, even if that's possible, the solution is still distinctively non-trivial.)

PR #39, opened 2023-06-05
Minimal changes to reduce contract size

proxima424:
## Context
Execution logic of modifier is added to each method's thereby increasing contract size.
It is a safer approach to inline the logic into a separate function.
0xFable, a game will benefit from saving on contract size wherever possible because of linear relation between
game logic and slocs.

## Describe Your Changes
Transferred the modifier's logic into two private functions namely `_checkIfExists()` and `checkLegitStep`

## Checklist

- [ X ] I have performed a self-review of my code
- [ X ] I ran `make lint` and fixed resulting issues
- [  ] I ran the relevant tests and they all pass

## Testing

Can run a codesize review after the all rest of the game logic is implemented.

### DIFF
--- a/packages/contracts/src/Game.sol
+++ b/packages/contracts/src/Game.sol
@@ -224,19 +224,22 @@ contract Game {
     // =============================================================================================
     // MODIFIERS
 
+
+
+    function _checkIfExists(uint256 _gameId) private {
+        if (gameData[_gameID].lastBlockNum == 0)
+            revert NoGameNoLife();
+    }
+
     // Check that the game exists (has been created and is not finished).
     modifier exists(uint256 gameID) {
-        if (gameData[gameID].lastBlockNum == 0)
-            revert NoGameNoLife();
+        _checkIfExists(gameId);
         _;
     }
 
     // ---------------------------------------------------------------------------------------------
 
-    // Checks that the game exists, that the requested step is compatible with the game state, and
-    // that the next player is the message sender.
-    modifier step(uint256 gameID, GameStep requestedStep) {
-
+    function _checkLegitStep(uint256 gameId, GameStep requestedStep) private {
         // NOTE(norswap): This whole function is pretty fragile and must be revisited whenever the
         //   the step structure changes.
 
@@ -295,6 +298,13 @@ contract Game {
         gdata.lastBlockNum = block.number;
     }
 
+    // Checks that the game exists, that the requested step is compatible with the game state, and
+    // that the next player is the message sender.
+    modifier step(uint256 gameID, GameStep requestedStep) {
+        _checkLegitStep(gameId,requestedStep);
+        _;
+    }
+
     // =============================================================================================
 
     constructor(Inventory inventory_, DrawVerifier drawVerifier_, PlayVerifier playVerifier_) {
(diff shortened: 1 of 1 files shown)


packages/contracts/src/Game.sol
```diff
@@ -224,19 +224,22 @@ contract Game {
...
+        if (gameData[_gameID].lastBlockNum == 0)
+            revert NoGameNoLife();
+    }
+
     // Check that the game exists (has been created and is not finished).
     modifier exists(uint256 gameID) {
-        if (gameData[gameID].lastBlockNum == 0)
-            revert NoGameNoLife();
+        _checkIfExists(gameId);
         _;
```
norswap, line 237: Does this save gas? Using very rough intuition, I'd expect it to only save gas if Solidity doesn't have very basic common sub-expression optimization... which is possible given a rather not great implementation.

(Let's find out! Can you run run both version with make test-gas in the contracts package?)

Even if it does save some gas, I don't think that one is worth the added clutter though.

packages/contracts/src/Game.sol
```diff
@@ -295,6 +298,13 @@ contract Game {
         gdata.lastBlockNum = block.number;
     }
 
+    // Checks that the game exists, that the requested step is compatible with the game state, and
+    // that the next player is the message sender.
+    modifier step(uint256 gameID, GameStep requestedStep) {
+        _checkLegitStep(gameId,requestedStep);
+        _;
+    }
+
```
norswap, line 307: This is a good idea ... but does it compile? You left `_;` in checkLegitStep! It needs to be split into two functions (one for before, and one for after). Probably a good idea to extract the gameData pointer in the step logic and pass it to both functions.

packages/contracts/src/Game.sol
```diff
@@ -312,7 +322,8 @@ contract Game {
     function fetchGameData(uint256 gameID) external view returns(FetchedGameData memory) {
         GameData storage gdata = gameData[gameID];
         PlayerData[] memory pData = new PlayerData[](gdata.players.length);
-        for (uint8 i = 0; i < gdata.players.length; ++i)
+        uint8 gdataLength = uint8(gdata.players.length);
+        for (uint8 i = 0; i < gdataLength; ++i)
```
norswap, line 326: Super curious to know whether this makes a difference actually!

It's another case where a moderately smart compiler should be able to figure it out (nothing in the loop has side effects other than writing `pData`, so `gdata.playerData` should never change, so it should be possible to figure out the length doesn't change).

Can you compare both versions to find out?

norswap (changes requested):

packages/contracts/src/Game.sol
proxima424: The main motive behind this was not to save gas but to ensure that some core functions are not overloaded and checks like this are forwarded to smaller helper functions and save contract size.
( Overloaded as in : All the modifier's code is inlined into the each function's code where the modifier's being used )

packages/contracts/src/Game.sol
proxima424: Yes, I'll compare both versions to find out, thanks :)
My motive was to save gas while calling SLOADs on the storage array for the length multiple times throught the iteration.
Will find out the gas savings.

packages/contracts/src/Game.sol
norswap: Yeah, this was a dumb comment on my part. I got the two things tangled in my head somehow and I was talking about size, but then assuming that it was being used only a single location, which is of course not the case.

norswap: Still think the size increase due to this is probably not worth making slightly less pleasant to the eye.

PR #44, opened 2023-06-08
Add circom info to readme

nezouse:
Adding a link to circom as i couldn't find how to install it before.


norswap:
Great, thanks!

PR #47, opened 2023-06-11
Initial hand refactor

eerkaijun:

This PR is an alternative to the other [PR](https://github.com/norswap/0xFable/pull/14), which is to verify the initial hand proof on-chain. It uses a different approach to the other PR, where in this PR, the player first has to commit a salt hash, and using that salt hash to generate randomness when drawing initial cards ([relevant issue](https://github.com/norswap/0xFable/issues/36)). 

## Describe Your Changes

Updated the `initial.circom` circuit to draw random indices in circuit instead of off-chain. Updated the corresponding verification contract and `Game.sol`. 

## Checklist

## Testing

Have updated the circuits and contracts test to make sure they work.


eerkaijun:
With this approach, I think the game flow might need to change a bit though. Currently, (if I understand correctly), a player can `joinGame` anytime, and during `joinGame`, the same transaction also means the player is drawing the initial cards. So essentially player1 can first join and draw cards, and player2 can come later. With committing salts, we want both players to first commit their respective salts, before any of them can join a game, I think?

eerkaijun:
@norswap since I previously merge `master` into `initial-hand-verification`, I think we can't rebase anymore due to non-linear commits. Seems like the repo disabled "Create a merge commit" option, which I think is the easiest option for this.

norswap (commented):
Looks like you reformatted all the contracts. Which to be fair, is the PR template, but quite obviously, we haven't really done before.

I'm actually not sure we want to impose the linting, I'm not a fan of some of the things it does and it's not really configurable. Need to look into it.

In any case, it would probably be better to remove the linting changes here, as it sorts of hide what changed in the contracts besides linting.

Other than that, not sure it's most judicious to spend my time doing a precise review here given that the contracts will need to change anyway and we did discuss the design at length already. But if you want me to look at something specific, let me know and I'll give it a look!

norswap:
I've:
- removed the linting (just dropped the linting commit)
- removed the changes to the lockfile (dropped the last "fix conflict" commit)
- relinearized everything, it was surprisingly easy, [here's how you do it](https://stackoverflow.com/a/57330304)

Pushed the result at [`initial-hand-refactor-rebased`](https://github.com/norswap/0xFable/compare/initial-hand-refactor-rebased). Let me know if it's all good, we can force-push it to this branch and merge that!

eerkaijun:
Okay just checked and tested it out, `initial-hand-refactor-rebased` looks good to me. Let's merge that then?

PR #56, opened 2023-06-20
Implement todo features in Inventory.sol

0xNONSO:
- Prevent cards from being added/removed to/from the inventory during a game
- Limit maximum number of card copies in a deck to three

Link related issues!
https://github.com/norswap/0xFable/blob/master/docs/architecture.md#cards-and-inventory

## Describe Your Changes
-  Added a modifier `notInGame()` to limit players from addding/removing cards to/from the inventory contract while participating in a game
- Modify `checkDeck()` to check if a player possesses more than three copies of  a card in a deck
## Checklist

## Testing

todo


packages/contracts/src/Inventory.sol
```diff
@@ -246,14 +266,21 @@ contract Inventory is Ownable {
...
+        for (uint256 i = 0; i < deckLength; ++i) {
             uint256 cardID = deck.cards[i];
             if (inventoryCardsCollection.ownerOf(cardID) != player)
                 revert CardNotInInventory(cardID);
+                
+            // check that each card does not exceed its maximum amount of copies
+            uint256 cardCopies;
+            for (uint256 j = 0; j < deckLength; ++j) {
+                if (cardID == deck.cards[j]) ++cardCopies;
+            }
```
norswap, line 280: This is not ideal, as you'll visit the whole deck for each card. So if the deck has size 64 (the max), you'll have 4096 iterations of this, not ideal.

It's also not correct I believe. Say that the deck contains 3 cards with id 1: `[1, 1, 1]` — the inner loop will increment `cardCopies` 3 times on the first iteration, and 3 times on each subsequent iteration, so `cardCopies`will end up with value 9, and not 3 as expected.

To be fair, other solutions suck as well: there is no built-in memory maps, and using a storage map would be too costly in gas (it would actually definitely be ok on Optimism today, and mayyyybe ok too on something like Gnosis chain, but it's not a future-proof solution).

Building a memory based map is possible, but it's a lot of trickiness just for this.

Another similarly fiddly approach: sort the array, then the same cards end up next to each other.
I think that might be the best approach tough.

Here is [some code](https://gist.github.com/subhodi/b3b86cc13ad2636420963e692a4d896f) for sorting an array in Solidity. Sorting is O(n log(n)) on average so for 64 cards, about 384 "iterations". I would benchmark it against [this implementation](https://gist.github.com/sdelvalle57/f5f65a31150ea9321f081630b416ed99) of bubble sort as well, as I'm not sure how much overhead the function call adds in the quicksort version.

packages/contracts/src/Inventory.sol
```diff
@@ -61,6 +68,9 @@ contract Inventory is Ownable {
     // Max number of cards in a deck.
     uint256 constant private MAX_DECK_SIZE = 40;
```
norswap, line 69: Let's bump this to 64, that's the value we've played with for the circuits.

packages/contracts/src/Inventory.sol
```diff
@@ -32,6 +32,13 @@ contract Inventory is Ownable {
     // A deck contains a card that the player hasn't transferred to the inventory.
     error CardNotInInventory(uint256 cardID);
 
+    // A player cannot add card, remove card, remove deck, replace deck, add card to deck, remove card from deck
+    // while participating in a game
+    error PlayerIsInActiveGame(address player);
+    
+    // Number of card copies in a deck cannot exceed `MAX_CARD_COPY`
```
norswap, line 39:
```suggestion
    // Number of card copies in a deck cannot exceed `MAX_CARD_COPY`.
```

packages/contracts/src/Inventory.sol
```diff
@@ -32,6 +32,13 @@ contract Inventory is Ownable {
     // A deck contains a card that the player hasn't transferred to the inventory.
     error CardNotInInventory(uint256 cardID);
 
+    // A player cannot add card, remove card, remove deck, replace deck, add card to deck, remove card from deck
+    // while participating in a game
```
norswap, line 36:
```suggestion
    // A player cannot add/remove cards to/from the inventory, nor modify decks, while participating
    // in a game.
```

norswap (changes requested):
Fantastic PR! Let's get the duplication code fixed and merge this!

Also some tests would be fantastic, I think just for the feature you're added. I know there is basically no testing discipline right now, but it's going to be easier if we start adding them with small changes like this.

0xnonso:
I added a very basic test for `checkDeck()` [here](https://github.com/0xnonso/0xFable/blob/master/packages/contracts/src/test/Inventory.t.sol). The test reverts fine if card copy is gt `MAX_CARD_COPY` but foundry's [expectRevert()](https://book.getfoundry.sh/cheatcodes/expect-revert) doesn't seem to catch it. 🥲
Might be an issue with the encoding if not then its probably a foundry bug.

packages/contracts/src/test/Integration.t.sol
```diff
@@ -65,6 +62,13 @@ contract Integration is Test {
         game.commitSalt(COMMITTED_SALT);
         game.joinGame(gameID, 0, bytes(""), HAND_ROOT, DECK_ROOT, PROOF);
         vm.stopPrank();
+    }
+
+    function testGame() public {
+        uint256 gameID = 1;
+        // TODO
+        //game.createGame(2, game.allowAnyPlayerAndDeck);
```
norswap, line 70: Let's delete these two lines (they were there because I had a "allowed to join game?" predicate that I removed because ethers.js couldn't easily encode it on the frontend, will read it later, but this TODO here is pretty useless).

packages/contracts/src/test/Inventory.t.sol
```diff
@@ -0,0 +1,61 @@
...
+    function setUp() public {
+        deployment = new Deploy();
+        deployment.dontLog();
+        deployment.run();
+
+        cardsCollection = deployment.cardsCollection();
+        inventory = deployment.inventory();
+        inventoryCardsCollection = deployment.inventoryCardsCollection();
+        game = deployment.game();
+
```
norswap, line 30:
```suggestion
```

packages/contracts/src/test/Inventory.t.sol
```diff
@@ -0,0 +1,61 @@
...
+        deployment.run();
+
+        cardsCollection = deployment.cardsCollection();
+        inventory = deployment.inventory();
+        inventoryCardsCollection = deployment.inventoryCardsCollection();
+        game = deployment.game();
+       
+    }
+
+    function testCheckDeck() public {
```
norswap, line 33: Would add a little comment here about what this is testing.

packages/contracts/src/Inventory.sol
```diff
@@ -8,8 +8,9 @@ import "./InventoryCardsCollection.sol";
 import "openzeppelin/access/Ownable.sol";
 import "./Game.sol";
 
-contract Inventory is Ownable {
+import {Utils} from "./lib/Utils.sol";
```
norswap, line 11: This file wasn't committed :D

packages/contracts/src/Inventory.sol
norswap, line 1: I would suggest rebasing this PR onto the branch here: https://github.com/norswap/0xFable/pull/57.
Otherwise, we'll get a lot of duplicated linting changes that will cause annoying merge conflicts.

Let me know if you want me to do it, I'm used to this kind of git-fu by now.

norswap (changes requested):

norswap:
Did you figure the testing issue? I had some time on the plane, so I coded up a bubble sort and tried testing with it, the test passed :)

Here's the bubble sort implementation, so you can benchmark which test implem is the most gas efficient:

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.0;

library Utils {
    function sort(uint256[] memory original) pure public returns (uint256[] memory) {
        if (original.length == 0) {
            return original;
        }
        uint256[] memory array = new uint256[](original.length);
        array[0] = original[0];

        // Each iteration inserts item `original[i]` into `array`, whose size is `i` and
        // will be expanded to size `i+1` as a result of the insertion.
        for (uint256 i = 1; i < array.length - 1; ++i) {
            uint256 j;
            // Each iteration will assign the value of `array[j]` based on the value of `array[j-1]
            // and `original[i]`.
            for (j = i; j > 0; --j) {
                if (original[i] > array[j-1]) {
                    array[j] = original[i];
                    break; // value inserted at j, no need to continue
                } else {
                    array[j] = array[j-1];
                }
            }
            // If `j == 0`, `original[i]` wasn't inserted yet
            if (j == 0) {
                array[0] = original[i];
            }
        }
        return array;
    }
}

```

0xnonso:
> Did you figure the testing issue? I had some time on the plane, so I coded up a bubble sort and tried testing with it, the test passed :)
> 
> Here's the bubble sort implementation, so you can benchmark which test implem is the most gas efficient:
> 
> ```solidity
> // SPDX-License-Identifier: UNLICENSED
> pragma solidity ^0.8.0;
> 
> library Utils {
>     function sort(uint256[] memory original) pure public returns (uint256[] memory) {
>         if (original.length == 0) {
>             return original;
>         }
>         uint256[] memory array = new uint256[](original.length);
>         array[0] = original[0];
> 
>         // Each iteration inserts item `original[i]` into `array`, whose size is `i` and
>         // will be expanded to size `i+1` as a result of the insertion.
>         for (uint256 i = 1; i < array.length - 1; ++i) {
>             uint256 j;
>             // Each iteration will assign the value of `array[j]` based on the value of `array[j-1]
>             // and `original[i]`.
>             for (j = i; j > 0; --j) {
>                 if (original[i] > array[j-1]) {
>                     array[j] = original[i];
>                     break; // value inserted at j, no need to continue
>                 } else {
>                     array[j] = array[j-1];
>                 }
>             }
>             // If `j == 0`, `original[i]` wasn't inserted yet
>             if (j == 0) {
>                 array[0] = original[i];
>             }
>         }
>         return array;
>     }
> }
> ```

yes yes, i figured the testing issues out. It was my fault, didn't realize that foundry was expecting the wrong call to revert!

norswap:
For easier reviewing, here's a diff view with the linting PR: https://github.com/norswap/0xFable/commit/eca7db7bd532d320bd9dd9d25629a4a256e46422

packages/contracts/src/Inventory.sol
```diff
@@ -245,22 +294,44 @@ contract Inventory is Ownable {
...
+                if (cardCopies > MAX_CARD_COPY){
+                    revert CardExceedsMaxCopy(cardID);
+                }
+            } else {
+                if (inventoryCardsCollection.ownerOf(cardID) != player){
+                    revert CardNotInInventory(cardID);
+                }
+                cardCopies = uint256(1); 
+            }
+            prev = sortedCards[i];
```
norswap, line 321: This can be in the `else` branch.

norswap (commented):
Looks pretty good to me!
Let me merge the lint PR and we can rebase this on top of master then merge.

Could you check the gas usage between the mergesort and the bubble sort?

norswap:
linting PR rebased onto master!

0xnonso:
> Looks pretty good to me! Let me merge the lint PR and we can rebase this on top of master then merge.
> 
> Could you check the gas usage between the mergesort and the bubble sort?

sure! seems there is a bug in the bubble sort implementation which always sets the last iteration to zero

0xnonso:
not decreasing the array length in the initial loop fixes it.
so code would look like:
`for (uint256 i = 1; i < array.length; ++i){ ... }`

0xnonso:
mergesort implementation is generally cheaper in most cases. 
this is a gas benchmark using a randomly generated array of length `48`.
```solidity 
uint256[] public randomArray = [9, 4, 12, 35, 22, 17, 46, 7, 28, 31, 19, 11, 42, 26, 38, 14, 2, 29, 1, 16, 23, 5, 33, 8, 45, 43, 15, 32, 18, 13, 10, 47, 20, 27, 3, 6, 41, 25, 40, 34, 21, 30, 37, 24, 36, 39, 44, 0];
```
sorting gas benchmark

packages/contracts/src/Inventory.sol
0xnonso: hmm, i dont understand.
if i put it in the `else` block then it would only be executed conditionally whereas it needs to be executed on each iteration.
what am i missing? 👀

norswap: It does not need to be executed in each iteration, only when the card changes. If you're in the `if` branch, then you know the card is the same as last one, so `prev` can remain the same. In the `else` branch is when the card changes and `prev`needs to be updated.

norswap:
> not decreasing the array length in the initial loop fixes it.

Yes, my bad!

> mergesort implementation is generally cheaper in most cases.
> this is a gas benchmark using a randomly generated array of length 48.

Super interesting, thanks for running the experiment!

norswap:
I missed something super fundamental in this: we're comparing card IDs, but card IDs are unique (they're NFT IDs). Instead, we need to compare some kind of "card type" identifier. Which we currently don't have, it being a demo and everything.

I think we should add a bogus `cardType` value in in the `Card` structure (can be an `uint32` for now, we can change later) and perform the check using this value.

packages/contracts/src/Inventory.sol
0xnonso: true true
got it!

0xnonso:
okok, currently inventory cards are based on the their `tokenId` so how would the `cardType` be determined?

norswap:
> okok. currently the identifier for inventory cards are based on their `tokenId` so how would the `cardType` be determined?

This is what I meant above:

> I think we should add a bogus cardType value in in the Card structure (can be an uint32 for now, we can change later) and perform the check using this value.

Basically, `tokenID` is an NFT token ID that can be used in the `CardCollection` to get the characteristics of the token (where token == unique card, just like a unique physical copy of a card in Magic)

Right now you can use this token to get the name of the card, it's stats, etc ... (the `getCard` function gets you all of that).

When we have multiple cards in decks at the moment we simply give them the same name and the same stats. This is a bit of a hack.

The proper solution would be to map each card (token ID) in `CardCollection` to a `cardTypeID`, then map that `cardTypeID` to names, stats, etc... In the future, cards may have some features that are not shared between all cards of the same kind. For instance, some cards may be shiny while others are plain, or some cards may get a special illustration, or they're from a different edition, ... but they still count as the same "card type" form a gameplay perspective.

The simplest hacky solution is just to add a `cardTypeID` field in the `Card` structure to make this work, but if you feel like it, you can go the extra mile and implement the proper system here.

I think it's probably preferable to just go with a hack and open a new issue + PR for the better system :)

PR #57, opened 2023-06-23
Improve Linting

norswap:

The Github issues asks for people to lint, but we haven't actually done this consistently. This PR intends to put the code in a linted state + improves the tooling around linting.

## Describe Your Changes

- Linted contract code
- Made sure compiled circuits could be `make install`ed from circuits repo and would work, and not case any undue warning or linting issue
     - The current verifier code was not consistent with the circom code in the repo! Fixed the code there to make things compile / let tests pass (circuit integration w/ contracts is untested atm).
     -  Avoiding linting problems required excluding the contracts from linting by moving them to a specific repo and specifying the other repos explicitly.
     - I add to ignore some warnings in solc (unused parameters), but `make lint` will still check for them (just not in Circom contracts).
- Removed the dependency on prettier in favour of using `forge fmt` to format code.

TODO:
- Similarly lint the frontend code.

## Checklist


norswap:
@eerkaijun Do you know why the existing contracts add `bytes calldata _proof` parameters, but those I'm generating with Circom (I have version 2.1.4) have `uint256[24] calldata _proof` instead?

eerkaijun:
@norswap Hmmm I just tried regenerating the verifier contract and it's giving me `bytes calldata _proof` as the parameters. I have 2.1.4 for circom too. What is your snarkjs version?

eerkaijun:
Okay seems like snarkjs recently just upgraded to 0.7.x version and I am using a slightly older version. Once I upgrade my snarkjs then I got the same parameters as you.

packages/contracts/src/Game.sol
```diff
@@ -496,30 +508,39 @@ contract Game {
...
     //
     // The player's deck is cards[pdata.deckStart:pdata.deckEnd].
-    function checkInitialHandProof(PlayerData storage pdata, bytes32 initialDeckRoot,
-             uint256 randomness, uint256 committedSalt, bytes calldata proof) view internal {
+    function checkInitialHandProof(
+        PlayerData storage pdata,
+        bytes32 initialDeckRoot,
+        uint256 randomness,
+        uint256 committedSalt,
+        bytes calldata proof
```
eerkaijun, line 516: With the new snarkjs version where the verifier contract takes in `uint256[24]` as proof, I'm wondering if the client side proof generation will also generate `BigInt[24]` instead of hex string? (Assuming we are using `snarkjs.plonk.fullProve(...)` on the client side?)

norswap: I'm guessing this might be the case indeed, will try it out when I get back to integrating proofs.

PR #60, opened 2023-07-27
Contract byte packing

eerkaijun:

Update smart contracts to fit with byte packing circuits

## Describe Your Changes

Updated verifier contracts and some minor game logic

## Checklist

- [/] I have performed a self-review of my code
- [/] I ran `make check` and fixed resulting issues
- [/] I ran the relevant tests and they all pass
- [/] I wrote tests for my new features, or added regression tests for the bug I fixed

## Testing

Made sure tests passed


packages/contracts/src/Game.sol
```diff
@@ -518,12 +519,14 @@ contract Game {
...
         // construct circuit public signals
-        uint256[68] memory pubSignals;
-        // TODO copy deck in slots 0-63
-        pubSignals[64] = uint256(initialDeckRoot);
-        pubSignals[65] = uint256(pdata.handRoot); // TODO should be supplied as argument
-        pubSignals[66] = committedSalt;
-        pubSignals[67] = randomness;
+        uint256[7] memory pubSignals;
+        pubSignals[0] = 0; // TODO: first field element
+        pubSignals[1] = 0; // TODO: second field element
```
eerkaijun, line 524: The only thing I'm not entirely sure is that should we parse the deck cards into field elements on the smart contract? Or should we refactor the representation of cards in the smart contract as field elements from the beginning?

packages/contracts/src/Game.sol
```diff
@@ -161,6 +161,7 @@ contract Game {
         uint16 health;
         uint8 deckStart;
         uint8 deckEnd;
+        uint8 handSize;
```
norswap, line 164: So the idea was that we could keep that private, but we'd have to check if `handSize == maxHandSize || maxSizedHand[handSize] == NO_CARD` in the proof, right?

Definitely not worth doing now, since cards can be counted anyway. In the future, not knowing the numbers of cards is kinda funny, but not super duper needed either.

packages/contracts/src/Game.sol
```diff
@@ -767,8 +783,11 @@ contract Game {
             revert CardIndexTooHigh();
         }
         uint256 card = gdata.cards[cardIndex];
-        checkPlayProof(pdata, handRoot, card, proof);
+        bytes32 committedSalt = salts[msg.sender];
```
norswap, line 786: The salts hashes should probably be stored in `PlayerData` but let's not change that here, I have it refactored locally, I can merge things.

norswap (commented):
Is there a way for you to test this easily? Like generating a proof for some dummy values and adding that in the integration tests?

We can use forge cheatsto read the proof from a disk file.

packages/contracts/src/Game.sol
```diff
@@ -537,6 +538,29 @@ contract Game {
         }
     }
 
+    // Helper function to pack cards into field elements.
+    function bytePacking(uint256[] memory deck) internal pure returns (uint256[2] memory packedCards) {
+        // pad the cards to 62 cards (255 for null value)
+        // 31 cards is packed into one single field element
+        uint256[] memory cards = new uint256[](62);
+        // TODO: we assume deck length is less than 62 (since MAX_DECK_SIZE is 40)
+        // but would be nice to have an additional check
```
norswap, line 547: Yeah, normally that will be checked upstream, we could add a `require(deck.length <= 62)` here.

packages/contracts/src/test/Integration.t.sol
```diff
@@ -57,7 +58,8 @@ contract Integration is Test {
         // TODO
         //game.createGame(2, game.allowAnyPlayerAndDeck);
         game.createGame(2);
-        inventory.getDeck(player1, 0);
+        inventory.getDeck(player2, 0); // player1 has card id of 49-72 inclusive
+        inventory.getDeck(player2, 0); // player2 has card id of 73-96 inclusive
```
norswap, line 62: Due to deploy script airdropping the two first anvil addresses. We probably shouldn't rely on this staying like this, but fine in the comments. Maybe good to add `// ids are not starting at 1 because the deploy script currently airdrops some addresses, might change`

eerkaijun:
- Made some changes to the `verify.js` script and can confirm that the generated input works with the verification key. 
- Updated the smart contract integration tests, can confirm that the proof verification passed. Currently only tests for the first proof verification (`checkInitialHandProof`).

packages/contracts/src/test/Integration.t.sol
```diff
@@ -57,14 +90,16 @@ contract Integration is Test {
...
+        // ids are not starting at 1 because the deploy script currently airdrops some addresses, might change
         vm.startPrank(player1);
         game.commitSalt(COMMITTED_SALT);
-        game.joinGame(gameID, 0, bytes(""), HAND_ROOT, DECK_ROOT, PROOF);
+        game.joinGame(gameID, 0, bytes(""), HAND_ROOT, DECK_ROOT, generateProof());
```
eerkaijun, line 98: Currently only test the first proof verification. When we set the `checkProof` variable to be true, this verification passes. Do you know if there is a foundry cheatcode that can change/set the value of a variable in the smart contract? (So that we can set the value of `checkProof` in the test)

Something like [this](https://github.com/defi-wonderland/smock) but for Foundry.

packages/contracts/src/test/Integration.t.sol
eerkaijun: Done!

packages/circuits/Makefile
```diff
@@ -2,7 +2,7 @@
 # VARIABLES
 
 # List of powersOfTau trusted setup files here: https://github.com/iden3/snarkjs/blob/master/README.md
-SETUP_FILE=powersOfTau28_hez_final_19.ptau
+SETUP_FILE=powersOfTau28_hez_final_18.ptau
```
norswap, line 5: How much size is the witness file to be downloaded on the frontend now? With the 10x reduction in constraints we achieve, couldn't we go into an even smaller file?

Does the powers of tau file actually affect witness file size? Or just the maximum number of constraint that can be handled?

packages/contracts/src/test/Integration.t.sol
norswap: There is load for setting a storage slot: https://book.getfoundry.sh/cheatcodes/load but you have to compute the position of the storage slot yourself.

Another option is add a function in Game.sol that checks the caller is 0x0 then toggles the checkProof value, then impersonate 0x0 with a cheatcode.

Yet another option is have checkProof be a parameter in the Game.sol constructor, and write a function in the deploy script that controls the value of that parameter.

I feel like the second option is convenient for now, not too much code and resistant to refactorings. Might be want to get rid of it in the future if we run into contract size issues (but maybe that's super unavoidable and we'll need to factor out Game.sol in multiple contracts anyway).

eerkaijun: Yeah the second option sounds good, let's go with that

packages/circuits/Makefile
eerkaijun: Yes, we can use a smaller file, just made the update.

As far as I know, the ptau file does not affect the witness file size, just the maximum number of constraints. But a smaller ptau means a smaller zkey file, which is still better on the frontend.

eerkaijun:
Ok this PR should be ready to go. The integration test just checks for the first zk proof verification. I think we don't need to check every proof verification in this integration test since that will mean introducing many hardcoded values.

packages/contracts/src/Game.sol
```diff
@@ -934,4 +934,10 @@ contract Game {
     }
 
     // ---------------------------------------------------------------------------------------------
+
+    // TODO: Placeholder function - to be removed in production
```
norswap, line 938: Imho I'd leave it production, not worth the trouble to remove.

norswap:
> Ok this PR should be ready to go. The integration test just checks for the first zk proof verification. I think we don't need to check every proof verification in this integration test since that will mean introducing many hardcoded values.

It might be good to check one instance of every type of proof imho, meant to see if the checkproofs functions are implemented correctly & everything works end-to-end.

But let's maybe merge this and keep this for another PR?

Fantastic work as always 🔥🔥

norswap (approved):

PR #66, opened 2023-08-12
Implement TODO tasks in inventory

0xNONSO:
- Prevent card deck from being modified in the inventory during a game
- Limit maximum number of card copies in a deck to three

Link related issues!
https://github.com/norswap/0xFable/blob/master/docs/architecture.md#cards-and-inventory

## Describe Your Changes
- Added a modifier `notInGame()` to prevent players from modifying card deck while participating in a game
- Implement `checkDeck()` to check if a player possesses more than three copies of  a card in a deck
## Checklist

## Testing
basic tests [here](https://github.com/0xNONSO/0xFable/blob/master/packages/contracts/src/test/Inventory.t.sol)

### DIFF
--- a/packages/contracts/src/CardsCollection.sol
+++ b/packages/contracts/src/CardsCollection.sol
@@ -22,6 +22,7 @@ struct Card {
     uint256 id;
     Lore lore;
     Stats stats;
+    uint32 cardTypeID;
 }
 
 error Unauthorized();
@@ -36,6 +37,7 @@ contract CardsCollection is ERC721, Ownable {
 
     mapping(uint256 => Lore) public lore;
     mapping(uint256 => Stats) private stats_;
+    mapping(uint256 => uint32) public cardType;
 
     // Airdrop manager.
     address public airdrop;
@@ -69,6 +71,8 @@ contract CardsCollection is ERC721, Ownable {
         _safeMint(to, tokenID, "");
         stats_[tokenID] = Stats(attack, defense);
         lore[tokenID] = Lore(name, flavor, URL);
+        // NOTE(nonso): `tokenID` serves as a placeholder for the cardType ID
+        cardType[tokenID] = uint32(tokenID);
     }
 
     function stats(uint256 card) external view returns (Stats memory) {
@@ -81,6 +85,10 @@ contract CardsCollection is ERC721, Ownable {
     }
 
     function getCard(uint256 card) external view returns (Card memory) {
-        return Card(card, lore[card], stats_[card]);
+        return Card(card, lore[card], stats_[card], cardType[card]);
+    }
+ 
+    function getCardType(uint256 card) external view returns(uint32){
+        return cardType[card];
     }
 }
--- a/packages/contracts/src/CardsCollection.sol
+++ b/packages/contracts/src/CardsCollection.sol
@@ -87,8 +87,4 @@ contract CardsCollection is ERC721, Ownable {
     function getCard(uint256 card) external view returns (Card memory) {
         return Card(card, lore[card], stats_[card], cardType[card]);
     }
- 
-    function getCardType(uint256 card) external view returns(uint32){
-        return cardType[card];
-    }
 }
--- a/packages/contracts/src/Game.sol
+++ b/packages/contracts/src/Game.sol
@@ -1035,7 +1035,7 @@ contract Game {
 
     // Returns true if a `player` is currently in an active game
     // This works because players cannot participate in multiple games simultaneously
-    function playerActive(address player) external view returns(bool){
+    function playerActive(address player) external view returns (bool) {
         return inGame[player] != 0;
     }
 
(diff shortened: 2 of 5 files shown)


packages/contracts/src/CardsCollection.sol
```diff
@@ -81,6 +85,10 @@ contract CardsCollection is ERC721, Ownable {
     }
 
     function getCard(uint256 card) external view returns (Card memory) {
-        return Card(card, lore[card], stats_[card]);
+        return Card(card, lore[card], stats_[card], cardType[card]);
+    }
+ 
+    function getCardType(uint256 card) external view returns(uint32){
+        return cardType[card];
     }
```
norswap, line 93: This is redundant if the `cardType` mapping is public: Solidity will generate a function called `cardType` that other contract can call that acts exactly like this.

It's necessary for `getLore` etc because Solidity is weird and the autogenerated function returns a tuple of values instead of the struct type (and you can't assign the tuple directly to the struct type, that would be way too easy).

packages/contracts/src/Game.sol
```diff
@@ -1032,4 +1032,12 @@ contract Game {
     }
 
     // ---------------------------------------------------------------------------------------------
+
+    // Returns true if a `player` is currently in an active game
+    // This works because players cannot participate in multiple games simultaneously
+    function playerActive(address player) external view returns(bool){
```
norswap, line 1038: This misses a space, you can run `make lint` in the contracts dir to lint (and `make check` to run other checks).

packages/contracts/src/Inventory.sol
```diff
@@ -151,7 +170,7 @@ contract Inventory is Ownable {
 
     // Transfers a card of the sender to the inventory, mints a soulbound inventory card to the
     // sender in return.
-    function addCard(address player, uint256 cardID) external delegated(player) {
+    function addCard(address player, uint256 cardID) external delegated(player) notInGame(player) {
```
norswap, line 173: Theoretically, this could be allowed, but it might be good to forbid it because there is really no reason to do this, and it might make frontend management a bit easier (no need to monitor these events while in-game).

But if we do this, shouldn't we also prevent `addDeck` by symmetry? I'm leaning yes for now.

packages/contracts/src/Inventory.sol
```diff
@@ -295,4 +347,15 @@ contract Inventory is Ownable {
     }
 
     // ---------------------------------------------------------------------------------------------
+
+    function getCardType(uint256[] memory cardIDArr) public view returns(uint256[] memory){
```
norswap, line 351: Would pluralized: `getCardTypes`

packages/contracts/src/test/Inventory.t.sol
```diff
@@ -0,0 +1,74 @@
...
+        vm.startPrank(cardsCollection.airdrop());
+        uint256 randomMint = cardsCollection.mint(player2, "Horrible Gremlin", "", "", 1, 1);
+
+        // add card `randomMint` to inventory.
+        changePrank(player2);
+        inventory.addCard(player2, randomMint);
+
+        uint8 deckId = 0;
+
+        // add card `randomMint` to player2's deck.
```
norswap, line 64: player 1's deck, right?

Also not sure it's necessary to add the card to player 2 to test this?

norswap (commented):
There we go, great work with a few details to change!

Sorry this took so long, kinda fell off my radar, don't hesitate to ping if I take more than a few days without showing signs of life!

packages/contracts/src/CardsCollection.sol
0xNONSO: right!

packages/contracts/src/Inventory.sol
0xNONSO: yes yes, you are correct.
no reason to limit this as it doesn't affect the game play

packages/contracts/src/test/Inventory.t.sol
0xNONSO: yes that was a mistake! its player 1's deck.
The idea is if a player adds say `card A` to the inventory contract, another player shouldn't be able to add `card A` to their deck and use it in a game. regardless it isn't necessary as it still tests if a player owns the corresponding inventory card. but for this scenario not adding the `card A` to the inventory first would make it invalid as there would be no corresponding inventory card to check against.

norswap: > but for this scenario not adding the card A to the inventory first would make it invalid as there would be no corresponding inventory card to check against.

Are you sure about this? I think `ownerOf` would return 0 and it would still revert for the good reason.

Could you in any case add a comment here saying we test this so that we are testing the likely case where the card is owned by the inventory contract but not assigned to the right player, instead of the less likely case where the card isn't owned by the inventory contract (we could also test both separately)?

norswap (approved):
Great! Add a comment to that test function and then ready to merge!

norswap:
🚀🚀🚀

PR #69, opened 2023-10-08
Account abstraction integration

eerkaijun:
Opening a draft PR as I go along since I'm a frontend noob and it's easier to discuss through the PR.

### DIFF
--- a/packages/webapp/src/utils/acounts.ts
+++ b/packages/webapp/src/utils/acounts.ts
@@ -1,0 +1,109 @@
+/**
+ * Utilities to deal with burner accounts.
+ *
+ * @module utils/accounts
+ */
+
+import {
+    createWalletClient,
+    http,
+    getContract,
+    concat,
+    encodeFunctionData,
+    toHex
+} from 'viem'
+import { mainnet } from 'viem/chains'
+import { generatePrivateKey, privateKeyToAccount } from 'viem/accounts'
+import { simpleAccountFactoryAbi, simpleAccountAbi } from 'src/utils/abis.ts'
+
+interface UserOperation {
+    sender: string
+    nonce: BigNumberish
+    initCode: BytesLike
+    callData: BytesLike
+    callGasLimit: BigNumberish
+    verificationGasLimit: BigNumberish
+    preVerificationGas: BigNumberish
+    maxFeePerGas: BigNumberish
+    maxPriorityFeePerGas: BigNumberish
+    paymasterAndData: BytesLike
+    signature: BytesLike
+}
+
+const paymasterAddress = process.env.PAYMASTER_ADDRESS as `0x${string}`
+const paymaster = getContract({
+    address: paymasterAddress,
+    abi: paymasterAbi
+})
+
+export async function createAccount() {
+
+    // use simple account factory to deploy a new smart account
+    // get the address of the smart account
+    // create sdk that reads private key from the browser storage and signs transactions
+
+    const privateKey = generatePrivateKey()
+    // store this private key in browser storage (is this secure?)
+    localStorage.setItem("privateKey", privateKey)
+
+    const account = privateKeyToAccount(privateKey)
+
+    // TODO: change chain
+    const client = createWalletClient({
+        account,
+        chain: mainnet,
+        transport: http()
+    })
+
+    // Generate initcode
+    const simpleAccountFactoryAddress = process.env.SIMPLE_ACCOUNT_FACTORY_ADDRESS as `0x${string}`
+    const simpleAccountFactory = getContract({
+        address: simpleAccountFactoryAddress,
+        abi: simpleAccountFactoryAbi
+    })
+    const initCode = concat([
+        simpleAccountFactoryAddress,
+        encodeFunctionData({
+            simpleAccountFactoryAbi,
+            functionName: 'createAccount',
+            args: [account.address, 0]
+        }) // set salt as 0
+    ])
+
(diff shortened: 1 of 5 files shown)


packages/webapp/src/utils/acounts.ts
```diff
@@ -0,0 +1,120 @@
...
+
+export async function createAccount() {
+
+    // use simple account factory to deploy a new smart account
+    // get the address of the smart account
+    // create sdk that reads private key from the browser storage and signs transactions
+
+    const privateKey = generatePrivateKey()
+    // store this private key in browser storage (is this secure?)
+    localStorage.setItem("privateKey", privateKey)
```
norswap, line 57: This should be probably added as an atom in the store, using `atomWithStorage` (just like we use for the gameID).

I don't think putting it in the store will give us a ton of benefits, but at least it keeps all the state (and all the persistent state) in one place, making it a little bit easier to reason about / explore the codebase.

packages/webapp/src/utils/acounts.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    const account = privateKeyToAccount(privateKey)
+
+    // TODO: change chain
+    const walletClient = createWalletClient({
+        account,
+        chain: localhost,
+        transport: http()
+    })
```
norswap, line 66: Same, `import { getWalletClient } from "wagmi/actions"`

packages/webapp/src/utils/acounts.ts
```diff
@@ -0,0 +1,120 @@
...
+        simpleAccountFactoryAddress,
+        encodeFunctionData({
+            abi: simpleAccountFactoryAbi,
+            functionName: 'createAccount',
+            args: [account.address, 0]
+        }) // set salt as 0
+    ])
+
+    // Generate calldata to execute a transaction
+    const simpleAccountAddress = await simpleAccountFactory.read.getAddress([account.address, 0]) as `0x${string}`
```
norswap, line 85: Using wagmi CLI (cf. prev comment) you can get a function in `generated.ts` that does this call without the need to piece to do the `getContract` call yourself.

(And same for all other usages of `getContract`, of course.)

packages/webapp/src/utils/acounts.ts
```diff
@@ -0,0 +1,120 @@
...
+    let userOp: UserOperation = {
+        sender: simpleAccountAddress,
+        nonce: Number(await paymaster.read.senderNonce([simpleAccountAddress])),
+        initCode: initCode,
+        callData: callData,
+        callGasLimit: 3_000_000, // hardcode it for now at a high value,
+        verificationGasLimit: 3_000_000, // hardcode it for now at a high value,
+        preVerificationGas: 2_000_000, // hardcode it for now at a high value,
+        maxFeePerGas: 2e9,
+        maxPriorityFeePerGas: 1e9,
```
norswap, line 109: Are those two definitive values unlike the ones with comments?

packages/webapp/src/utils/acounts.ts
```diff
@@ -0,0 +1,120 @@
...
+        paymasterAndData: concat([
+            paymasterAddress,
+            `0x${'00'.repeat(64)}`,
+            `0x${'00'.repeat(65)}`
+        ]) as `0x${string}`,
+        signature: '0x' + '00'.repeat(65) as `0x${string}`
+    }
+
```
norswap, line 117: We should investigate if [permissionless.js](https://docs.pimlico.io/permissionless/reference) (integrated with viem) makes the flow here easier.

packages/webapp/src/utils/acounts.ts
eerkaijun: In there a way for us to pass in the `account` object in that case?

norswap: I think we should setup so that whenever we use a burner key, we setup the wagmi wallet with it. So this would mean moving the account creation / insertion to `src/wagmi.ts` and gating it with an option (I guess for now just a file-level boolean is fine).

norswap: You're right. I think we need to figure out a way to pass this client to wagmi however. i.e. when we use a burner key, we eschew the walletconnect prompting etc, and instead configure the global wagmi wallet to use the burner key.

norswap: Maybe the move is to create a custom connector in wagmi. This is poorly documented. Maybe the MockConnector suffices? https://wagmi.sh/core/connectors/mock

I'll try having a look soon.

norswap:
I think we should strongly consider making sure everything is plugged into a viem walletClient (which we then inject in wagmi) so that everything works transparently.

... if that's possible. Not sure if viem makes assumptions about the format of transactions however or if the interfaces are generic enough to allow this.

It's not clear to me that [permissionless.js](https://docs.pimlico.io/permissionless/reference/clients/bundlerClient) (which is viem-integrated) allows this directly. I'll inquire.

packages/webapp/src/utils/acounts.ts
norswap: Example for creating connectors: https://wagmi.sh/examples/custom-connector
Connector interface: https://github.com/wagmi-dev/wagmi/blob/main/packages/connectors/src/base.ts

eerkaijun:
Ok this PR is very outdated, will close this off. Making the integrations in a new branch (with permissionlessjs) https://github.com/0xFableOrg/0xFable/tree/account-abstraction

Issue #25, opened 2023-05-31
Implement a horizontal card drawer component


norswap:
This component would display a list of cards side by side, centered vertically when not taking up the whole screen.

This component can be used for displaying the cards in our hand, the cards in play under our control, and the cards in play under the opponent's control.

There should be some kind of effect when hovering a card (glow around the border, slight zoom), and a click on a card would expand the card detail (just a smaller version of the card picture is displayed in this "drawer"). Probably a good idea to include attack and defense for the creatures too.

If there are too many cards to display in the space available, there should be arrows icons on the left and right of the drawer to enable scrolling. If at all possible, keeping the arrow clicked should smooth scroll through the cards in the desired direction (to be tested for feel). We should also enable the mouse wheel to scroll in the selected area, maybe the arrow keys also (though maybe we should reserve this for keyboard card navigation?). Ideally, drag and drop in empty spaces should also enable us to to pan the drawer.

norswap:
![Untitled](https://github.com/norswap/0xFable/assets/202175/50a098d4-8b18-4410-a352-6af3f42a188c)

Very ugly wireframe to give some intuition of what the MVP is like here

PR #70, opened 2023-10-09
Drawer UI

ultraviolet10:

[Link related issues!](https://github.com/0xFableOrg/0xFable/issues/25)

## Describe Your Changes
Added CSS changes to `Hand`, `Card` components to resemble card drawer.
Made changes to scroll hooks, added arrows that use the same.

## Checklist

## Testing

### DIFF
--- a/packages/webapp/src/components/card.tsx
+++ b/packages/webapp/src/components/card.tsx
@@ -1,5 +1,5 @@
-import { useState } from "react";
-import Image from "next/image";
+import { useState } from "react"
+import Image from "next/image"
 
 // quick fix for hackathon
 const cards = [
@@ -94,13 +94,13 @@ export const Card = ({
   className,
   handHovered,
 }: {
-  id: number;
-  className?: string;
-  handHovered?: boolean;
+  id: number
+  className?: string
+  handHovered?: boolean
 }) => {
   // const [ , addToBoard ] = useAtom(store.addToBoard)
-  const [isDetailsVisible, setIsDetailsVisible] = useState<boolean>(false);
-  const [cardHover, setCardHover] = useState<boolean>(false);
+  const [isDetailsVisible, setIsDetailsVisible] = useState<boolean>(false)
+  const [cardHover, setCardHover] = useState<boolean>(false)
   return (
     <div
       className={`${className} transition-all ease-in-out duration-300 ${
@@ -122,8 +122,8 @@ export const Card = ({
           isDetailsVisible
             ? "font-serif overflow-hidden text-2xl font-bold text-slate-200 text-left text-clip max-w-[320px] select-none"
             : handHovered
-              ? "font-serif text-[14px] font-bold text-slate-200 text-left truncate select-none"
-              : "hidden"
+            ? "font-serif text-[14px] font-bold text-slate-200 text-left truncate select-none"
+            : "hidden"
         }
       >
         {cards[id]?.name}
--- a/packages/webapp/src/components/card.tsx
+++ b/packages/webapp/src/components/card.tsx
@@ -108,7 +108,6 @@ export const Card = ({
           ? "shadow-2xl z-[50] flex h-[33rem] max-w-[24rem] scale-[65%] cursor-pointer flex-col items-center justify-evenly rounded-lg border-4 bg-gray-900 p-5 transform translateY-[-50%]"
           : "flex flex-col space-y-1 max-w-[200px]"
       }`}
-      onMouseDown={(e) => e.preventDefault()}
       onClick={
         () => {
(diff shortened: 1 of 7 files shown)


packages/webapp/src/hooks/useScrollBox.ts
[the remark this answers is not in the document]
norswap: Some (maybe related, maybe not) scrolling bugs:

- Using the scrollbar to scroll doesn't seem to save the X position, so the position resets to X = 0, whenever the mouse exits the scrollbar.
- Moving laterally after clicking card details causes a violent scrolling motion towards the side opposing the mouse motion. Doesn't *always* happens but happens often.

norswap, line 1: On Chrome the cards are left-aligned (scrollbar left, last cards not visible), on Firefox, the cards appears right-aligned (scrollbar on the right, first cards not visible). It would be good to rationalize this.

(To be clear: the order of the cards is the same, it's just that the scrollbox is fully scrolled right on Firefox).

packages/webapp/src/components/card.tsx
norswap, line 1: I would try to make it so that card names and images cannot be selected, as that sometimes happens and looks janky.

packages/webapp/src/components/card.tsx
```diff
@@ -88,32 +89,70 @@ const cards = [
...
+      <span
+        className={
+          isDetailsVisible
+            ? "font-serif overflow-hidden text-2xl font-bold text-slate-200 text-left text-clip max-w-[320px]"
+            : hover
+            ? "font-serif text-[14px] font-bold text-slate-200 text-left truncate"
+            : "hidden"
+        }
+      >
+        {cards[id]?.name}
```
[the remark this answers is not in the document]
ultraviolet10: This was added since some card IDs were sent in that weren't part of the hardcoded `cards` list (in `card.tsx`).

packages/webapp/src/pages/play.tsx
```diff
@@ -32,30 +42,34 @@ const Play: NextPage = () => {
...
+        <GameEndedModal closeCallback={() => setHideResults(true)} />
+      )}
 
       <main className="flex min-h-screen flex-col">
         <Navbar />
 
         <Hand
           cards={playerHand}
-          className="mt-500 absolute z-[100] translate-y-1/2 transition-all duration-500 ease-in-out hover:translate-y-0"
+          className="absolute left-0 right-0 mx-auto z-[100] translate-y-1/2 transition-all duration-500 rounded-xl ease-in-out hover:translate-y-0"
```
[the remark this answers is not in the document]
ultraviolet10: Also getting `PrivateInfo` from store, populating `PlayerHand` which is fed into `Hand`.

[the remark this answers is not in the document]
norswap: Oh I see! I think for me it either errored or didn't show any card because I had to set the `cards` manually. Which can be fine for testing, instead of pulling the deck in.

packages/webapp/src/hooks/useScrollBox.ts
```diff
@@ -1,102 +1,91 @@
...
+  const cardWidth = 200 // width of card when not in focus
+  const scrollAmount = 2 * cardWidth
+  const duration = 300
+
+  const checkArrowsVisibility = () => {
+    if (!scrollRef.current) return
+    const { scrollLeft, scrollWidth, clientWidth } = scrollRef.current
+    setShowLeftArrow(scrollLeft > 0)
+    setShowRightArrow(scrollLeft < scrollWidth - clientWidth)
+  }
```
norswap, line 23: This needs to be called when the window resizes, otherwise if the arrows are not visible (e.g. window full screen) but you resize, then the arrows are not available anymore.

You can use `const playerHand = [0n, 1n, 2n, 3n, 4n, 5n, 6n, 7n, 8n, 9n]` to test (adapt the number of cards to your screen size!)

norswap (approved):
No more jankyness!

Just the resize issue to fix.

I'd also like to add the possibility of scrolling with the mouse wheel, **if it's not too hard**. Could be in this PR, or a subsequent PR, up to you! If not in this PR, fine to merge when the resize issue is fixed ✨

packages/webapp/src/hooks/useScrollBox.ts
ultraviolet10: Arrows now show up in the hand drawer when the window is resized.

ultraviolet10: Support for scrolling with mouse wheel added, to be tested.

PR #71, opened 2023-10-15
Draw Initial Hand Proof Integration

norswap:
Losing my absolute sanity on this.

The problem: I can generate a zk proof for drawing the initial hand locally, I can also verify it locally.
However, on-chain the verification fails systematically.

I've setup an event that relays all the public inputs and the proof used on-chain. They are absolutely identical to the ones used off-chain.

It's to the point where I'm seriously suspecting some kind of snarkjs-side bug.


norswap:
Fixed — I was sampling random values from [0, 256) instead of bounding them to the field prime. There probably is a bug in Snarkjs that makes it possible to prove & verify with values that are too large: https://github.com/iden3/snarkjs/issues/436

PR #72, opened 2023-10-16
fix field prime

kevincharm:

I believe the BN254 field prime/curve order should be `21888242871839275222246405745257275088548364400416034343698204186575808495617`

circom docs: https://docs.circom.io/circom-language/basic-operators/
aztec: https://github.com/noir-lang/noir/blob/master/noir_stdlib/src/compat.nr

## Checklist

## Testing

Some integration tests fail for me (even without the changes).


norswap:
Uh, interesting!

I got mine from here: https://hackmd.io/@jpw/bn254
Is that incorrect or did misunderstand something about that article?

kevincharm:
If you scroll down that page, it goes on to show the curve order, which is the value at which the field wraps around to 0!

norswap:
That makes sense, thanks! I'll rename the constant to CURVE_ORDER then!

Issue #74, opened 2023-10-21
Switch to Groth16


norswap:
Currently we use Plonk, which is nice because it doesn't require a circuit-specific trusted setup.

However, after all our optimizations, the DrawHand proof is still ~22s of client-side proving, and that's on a maxed-out M1 Mac. That's probably too slow for comfort, and it's a fact that the Groth16 implementation in snarkjs should be a lot faster.

The change should be relatively trivial — just change the Makefile & various calls to snarkjs to reflect Groth16 usage instead of Plonk. The major change might the need to add the generation of the circuit-specific trusted setup to the zk package Makefile.

When going to prod, we'll need to run a credibly decentralized trusted ceremony, we can use [p0tion](https://github.com/privacy-scaling-explorations/p0tion) from PSE for that prupose.

norswap:
Completed via #87, we'll need to figure out the trusted ceremony, but that's way out.

PR #87, opened 2023-12-17
Groth16

eerkaijun:

Link to issue: https://github.com/0xFableOrg/0xFable/issues/74

Switch ZK proving system from Plonk to Groth16 - major improvement in client side proving time. DrawInitalHand circuit proving reduced from 20s to 1.5s!

## Describe Your Changes

Main changes are:
- Update circuits proving key generation
- Update Solidity verifiers
- Update frontend to use Groth16
- Parsing the proof since Groth16 proof has a different format than Plonk

An interesting observation is that when using `snarkjs.groth16.fullProve` to generate the proof, it uses a different Endian that what the Solidity verifier would take it, so when parsing the proof we have to switch the Endian. Spent so long trying to figure this out LOL and have to dig through their source code.

## Checklist

## Testing

Updated some circuit tests and also tested on frontend.


packages/webapp/src/utils/zkproofs/proofs.ts
```diff
@@ -79,12 +86,16 @@ export async function prove
...
-
-    // Clean this fucking mess up.
-    return { proof: JSON.parse(calldata.split("][", 1)[0] + "]").map(BigInt) }
+    // Parse proof outputs
+    const generated_proof: ProofOutput =  { 
+      proof_a: proof["pi_a"].slice(0,2) as [bigint, bigint],
+      proof_b: [
+        [proof["pi_b"][0][1], proof["pi_b"][0][0]],
+        [proof["pi_b"][1][1], proof["pi_b"][1][0]]
+      ] as [[bigint, bigint], [bigint, bigint]],
```
eerkaijun, line 95: This is the Endian difference that needs to be parsed in order to match the solidity verifier.

norswap: Let's add a comment to that affect in the code.

norswap:
Amazing! How does it work wrt to the trusted setup? I thought groth required a circuit-specific setup, but we're still just using the powers of tau here?

eerkaijun:
> Amazing! How does it work wrt to the trusted setup? I thought groth required a circuit-specific setup, but we're still just using the powers of tau here?

The zkey that is used will be the one involved in the trusted ceremony. Currently in the `Makefile` we just use a zkey that has no contribution. In production, the final zkey that is used will be the output of the trusted ceremony with contributions who those that participated in the trusted ceremony.

Issue #92, opened 2023-12-17
Split Game.sol to fit under the contract size


norswap:
Currently `Game.sol` is way over the EVM contract size limit.

The simplest way to split things up is to use external libraries [as explained here](https://twitter.com/alphaK3Y/status/1729265795434451123?t=tsSRAWtudovyhHXOiFTljw).

PR #98, opened 2024-01-07
refactor: split Game.sol to fit under contract size

0xnonso:
Game.sol contract code size exceeds max limit. closes https://github.com/0xFableOrg/0xFable/issues/92.

## Describe Your Changes
Split Game.sol to fit under contract size.

## Checklist

## Testing
nil


packages/contracts/src/libraries/GameErrorsLib.sol
norswap, line 1: Given the relatively tight scope of the current codebase, I'd rename:
-  `GameErrorsLib` → `Errors`
- `GameEventsLib` → `Events`
- `GameStepLibs` → `GameActions`
- `GameStructsAndEnums` → `Structs`
- `GameUtils` → `Utils` (as mentioned somewhere else, they're not even game specific!)

packages/contracts/src/libraries/GameErrorsLib.sol
```diff
@@ -0,0 +1,105 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+pragma solidity ^0.8.0;
+
+library GameErrorsLib {
+    // ERRORS
+
+    // TODO This is an error from the inventory contract, but we need to reproduce it here
+    //      so that it is included in the ABI, which allows the frontend to parse the error.
+    //      We need to figure out a way to systematically import all errors that we may revert with
+    //      from other contracts.
```
norswap, line 10: Did you check to see if this problem is present after separating out the files? If not I'll have a look on how to check.

packages/contracts/src/libraries/GameInternalLib.sol
norswap, line 1: If we moved all of those to `GameStepLib`, would we still have a size problem?

Note that if we additionally move `concedeGame` and `timeout` form `Game` to `GameStepLib`, then the only function here that is used in both `Game` and `GameStepLib` is `nextPlayer`. The other three functions could be made `private` in `GameStepLib`.

For `nextPlayer`,  we could probably refactor `playerDefeated` so that we defer updating the player list until the end of `step`  — then `nextPlayer` can live as a function only in `Game` (but let's not do that now — maybe another PR).

packages/contracts/src/Game.sol
norswap, line 1: You should modify the Makefile to remove the size extension in the `anvil` target.

packages/contracts/src/libraries/Constants.sol
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+library Constants {
+    // =============================================================================================
+    // CONSTANTS
```
norswap, line 6: We can remove that comment and separator :)

packages/contracts/src/libraries/GameErrorsLib.sol
```diff
@@ -0,0 +1,105 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+pragma solidity ^0.8.0;
+
+library GameErrorsLib {
+    // ERRORS
```
norswap, line 5: Remove comment.

packages/contracts/src/libraries/GameEventsLib.sol
```diff
@@ -0,0 +1,68 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+pragma solidity ^0.8.0;
+
+library GameEventsLib {
+    // =============================================================================================
+    // EVENTS
```
norswap, line 6: Remove comment and separator.

packages/contracts/src/libraries/GameInternalLib.sol
norswap: Investigating a bit more, what seems to happen by looking at `broadcast/Deploy.s.sol/1337/run-latest.json` is that the solidity compiler notices that all libraries are being imported in `GameStepLib` and so basically only deploys that ones, every reference to any library will end up being a reference to that contract.

So yep, let's go ahead and merge `GameInternalLib` with `GameStepLib`, it'll more closely reflect the structure of the contracts + they're pretty fundamentally tied as implementing the same functionality.

norswap (changes requested):
Nice job! Few things I'd like to change, but this is going the right direction!

packages/contracts/src/libraries/GameInternalLib.sol
0xnonso: > Investigating a bit more, what seems to happen by looking at `broadcast/Deploy.s.sol/1337/run-latest.json` is that the solidity compiler notices that all libraries are being imported in `GameStepLib` and so basically only deploys that ones, every reference to any library will end up being a reference to that contract.
> 
> So yep, let's go ahead and merge `GameInternalLib` with `GameStepLib`, it'll more closely reflect the structure of the contracts + they're pretty fundamentally tied as implementing the same functionality.

yes, libraries with internal functions are appended to the contract's bytecode. so technically, `GameInternalLib.sol` is duplicated into both `GameStepLib.sol` and `Game.sol`. My reason for doing that is to prevent ever having an extra delegatecall  in the step modifier function when calling `playerDefeated(...)`

0xnonso: Also, regarding the libs my thoughts are related functions should be grouped together into separate libs i.e all game steps which include `play`, `attack`, `defend` go into `GameStepLib.sol` then all game actions which include `createGame`, `joinGame`, `cancelGame`, `concedeGame` etc go into `GameActionLib.sol`. wdyt?

norswap: > yes, libraries with internal functions are appended to the contract's bytecode.

Interesting, does that also affect non-internal functions?

e.g. if a have a library with both internal and non-internal functions, everything is always appended?

OR

the internal functions are always appended, but the non-internal functions are always deployed to a separate contract?

norswap: > Also, regarding the libs my thoughts are related functions should be grouped together into separate libs i.e all game steps which include `play`, `attack`, `defend` go into `GameStepLib.sol` then all game actions which include `createGame`, `joinGame`, `cancelGame`, `concedeGame` etc go into `GameActionLib.sol`. wdyt?

There isn't really such a distinction between "actions" and "steps", but I'm guessing you're basing it on whatever gets the "step" modifier right? (In the frontend I call "actions" that result logically from one user interaction, so `joinGame` + `drawInitialHand` are one action, `attack` is an action, etc...)

I do think a split on before joining / after joining makes sense, which is almost this. But `concedeGame` can only be called after joining (and drawing the initial hand), so theoretically it should be along the steps too.

That's technically not the case of `timeout` who doesn't really fit this as it can both be called before and after the game starts (drawing the initial hand can timeout). But since it makes more sense from the point of view of file organization, let's put it with the in-game actions.

But I'd still call the library that is currently called `GameStepsLib` to use the word "action". Right now, "steps" in the contract is a "state". i.e. in the `PLAY` step, a player can play a card OR attack. Let's call it `InGameActions`? And let's add a comment to `timeout` to say it can be called before the game starts too.

0xnonso: > > yes, libraries with internal functions are appended to the contract's bytecode.
> 
> Interesting, does that also affect non-internal functions?
> 
> e.g. if a have a library with both internal and non-internal functions, everything is always appended?
> 
> OR
> 
> the internal functions are always appended, but the non-internal functions are always deployed to a separate contract?

yes, it does. if the library contains both internal and non-internal functions it is deployed separately. it is only appended if the lib contains only internal functions

packages/contracts/src/libraries/GameErrorsLib.sol
0xnonso: The error is included in the generated ABI for the inventory contract. I assuming you mean the auto generated ABI in `generated.ts`.

packages/contracts/src/libraries/GameActionLib.sol
```diff
@@ -0,0 +1,353 @@
...
+pragma solidity ^0.8.0;
+
+import {CardsCollection, Stats} from "../CardsCollection.sol";
+import {Events} from "./Events.sol";
+import {Errors} from "./Errors.sol";
+import {GameStep, PlayerData, GameData} from "./Structs.sol";
+import {Constants} from "./Constants.sol";
+import {Utils} from "./Utils.sol";
+
+library GameActionLib {
```
norswap, line 11: Add a comment separator here.

norswap: And remove `Lib` from lib name & file name.

norswap (approved):
Nice work!

norswap:
You need to rebase this onto master locally.

Here's how you can do it: 

```
# this will remove your merge commit and your revert to the equivalent state without these commits
git reset --hard head~2

git remote add upstream git@github.com:0xFableOrg/0xFable.git
# or if using https: git remote add upstream https://github.com/0xFableOrg/0xFable.git
git fetch upstream
git rebase origin/master
git push --force
```

norswap:
🚀

PR #101, opened 2024-01-12
hotfix for card art not loading

eviterin:

Seems we are in drawInitialHand.ts assume that there are cards up to index 255.
When checking card-list.ts we only have 12 cards.
This means that we often get cards without any art/info when playing the game.

This is a simple fix, which just adds cards up to index 255.

## Checklist

- [X] I have performed a self-review of my code

## Testing
Opened two instances of the game.
Updated the card-list.ts file with my update.
Asserted that cards were loading.


norswap:
Thanks!

Issue #102, opened 2024-01-14
Merge Collection and Editor


eviterin:
Both the Collection and Editor views are similar in terms of design and functionality. They both feature a search and filter section on the left, a display of cards in the center, and a column to the right. The right column differs between the two: in Collection View, it shows all the player's decks, while in the Editor, it displays the cards in the deck being built.

Right now, we're redirecting the player between the two pages. 
From Collection to Editor when either "New Deck" or an existing deck is clicked. 
From Editor to Collection when either the deck save or discard button is clicked.

Instead of redirecting the player, we should just have the right column change. 

This is done when:
* The editor is merged into the collection file.
* When the deck list has been refactored into its own component.
* When the card list has been refactored into its own component.

eviterin:
@norswap could you please review and label?

eviterin:
Done as of PR #108

eviterin:
Done as of PR #108

Issue #93, opened 2023-12-17
Overhaul card collection display


norswap:
TODO

norswap:
We should tackle these items while doing this PR: https://github.com/0xFableOrg/0xFable/pull/108#issuecomment-1974002751

PR #108, opened 2024-01-20
The (Editor and Collection) Merge

eviterin:
Editor was added in my previous PR. This editor was based on the Collection view, and shared a lot of component and functionality. Instead of maintaining the two, I now combined them.

[Merge Collection and Editor](https://github.com/0xFableOrg/0xFable/issues/102)

## Describe Your Changes
Both Collection and Editor had these two components:
On the left: Search/Filter panel
Middle: Card collection 

In Collection:
On the right: Player's completed decks

In Editor:
On the right: Player's deck that is under construction

Each of these components have been factored out into their own component. 
Editor has been removed, and the Collection now keeps track on whether the player is in editing mode or not.
Index file has been updated to remove the editor shortcut.

Additionally, variables within the collection view have been reordered and grouped by the functionality they support.

## Checklist

## Testing
Exploration testing to make sure there are is no regression within the Collection view.


norswap:
Is this ready for review? Just tried it and seems to have quite a few problems (e.g. deck doesn't appear in list when you add a new deck) + unused functions in `collections.tsx`.

eviterin:
I ran eslint and resolved all warnings. Sorry for that will try to do better in the future.. 

@norswap I tried cloning my PR fresh and still cannot reproduce the issue you're having with decks not appearing in the list.
The following steps I take:
1. Navigate to http://localhost:3000/?index=1
2. Select Collection
3. Select New Deck
4. Add cards and a name and press Save
5. Assert that the deck I created is on the deck list.

Keeping this issue in draft until I know more.

norswap:
Might be on my side, maybe I need to rerun `make setup`, will report back!

norswap (commented):
Posting some comments I'd enqueued — not a complete review

packages/webapp/src/components/collection/cardCollectionDisplay.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+      <div className="col-span-7 flex rounded-xl border overflow-y-auto">
+        {isHydrated && cards.length === 0 && (
+          <div className="flex flex-row w-full justify-center items-center">
+            <MintDeckModal callback={refetch} />
+          </div>
+        )}
+
+        {isHydrated && cards.length > 0 && (
+          <div className="grid grid-cols-4 overflow-y-auto pb-4">
```
norswap, line 31: I know this was there before, but the `grid-cols-4` + the rest means that as we resize the window all cards resize, up to very small sizes

It might be good to change that to simply display as much as we need.

Okay, after much trial & error, querying chat GPT and the tailwind / CSS docs:

- Replace the grid line with 
`<div className="flex flex-wrap justify-around overflow-y-auto pb-4">`
- Add `grow w-[220px] max-w-[330px]` to the card div.
- Add `className="aspect-square"` to the image
- Remove `style={{ height: 'fit-content' }}`
    
This causes elements to grow dynamically from 220 to 300px, and if they reached max size, the remaining space is put around the elements in a balanced way (`justify-around`). The `aspect-square` thing simply replaces the `fit-content` line in a way that is clearer.

The image stuff ensures the image resizes

norswap (commented):
Will continue reviewing tomorrow, here's a first batch of comments.

packages/webapp/src/components/collection/deckCollectionDisplay.tsx
norswap, line 1: Let's call the file `deckList` like the component. Same for `deckPanel`.

packages/webapp/src/components/collection/deckConstructionPanel.tsx
```diff
@@ -0,0 +1,88 @@
...
+    }
+  
+    const handleSave = () => {
+      if (!deckName.trim()) {
+        setIsDeckNameValid(false)
+        return
+      }
+
+      const newDeck = {
+        name: deckName,
```
norswap, line 30: Let's also `trim` here.

packages/webapp/src/components/collection/deckConstructionPanel.tsx
```diff
@@ -0,0 +1,88 @@
...
+  }
+  
+  
+  const DeckConstructionPanel : React.FC<DeckConstructionPanelProps> = ({ deck, selectedCards = [], onCardSelect, onSave, onCancel }) => {
+    const [ deckName, setDeckName ] = useState(deck.name)
+    const [ isDeckNameValid, setIsDeckNameValid ] = useState(true)
+
+    const handleDeckNameChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+      setDeckName(event.target.value)
+      setIsDeckNameValid(true)
```
norswap, line 20: We should probably check if the trimmed value is valid. Otherwise we can type only spaces and it seems like the deck name is valid.

packages/webapp/src/components/collection/deckConstructionPanel.tsx
```diff
@@ -0,0 +1,88 @@
...
+    </div>
+  )
+}
+
+DeckConstructionPanel.defaultProps = {
+    deck: {
+      name: 'New Deck',
+      cards: []
+    }
+  }
```
norswap, line 86: I feel like this isn't used? Since by default the New Deck button leaves the deck name field empty.

packages/webapp/src/components/collection/deckConstructionPanel.tsx
```diff
@@ -0,0 +1,88 @@
...
+      {/* List of Cards in the Deck */}
+      <div className="mt-4 w-full">
+        {selectedCards.map((card, index) => (
+          <div 
+            key={index} 
+            className="p-2 border-b last:border-b-0 cursor-pointer hover:bg-gray-100"
+            onClick={() => onCardSelect(card)}
+          >
+            <div className="flex items-center space-x-3">
+              {/* todo @eviterin: get proper link to the card instead of always the witch */}
```
norswap, line 70: This is probably feasible now, right?

norswap (commented):
Nice work! Let's get the comments sorted and we can merge this :)

eviterin:
Minor CSS issues now with Aritra's changes, but will be fixed in upcoming PR for onchaining decks.

packages/webapp/src/components/collection/cardCollectionDisplay.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+const CardCollectionDisplay: React.FC<CardCollectionDisplayProps> = ({ cards, isHydrated, setSelectedCard, selectedCards, onCardToggle, isEditing }) => {
+  return (
+    <>
+      <div className="col-span-7 flex rounded-xl border overflow-y-auto">
+        {isHydrated && cards.length === 0 && (
+          <div className="flex flex-row w-full justify-center items-center">
+            <MintDeckModal/>
+          </div>
```
norswap, line 22: Whenever clicking some of the filter items in the left panel, this is shown. Clearly not what we want (because the filtered set size is 0).

The condition here should be changed to display this modal only if the full card collection (not the filtered one) is empty.

Btw, I want to refactor the left panel according to this later: https://github.com/0xFableOrg/0xFable/issues/93 and remove these filter buttons. But I think it's good to keep the generic filtering ability. This (refactor left panel) is for a further PR!

packages/webapp/src/components/collection/deckPanel.tsx
```diff
@@ -0,0 +1,84 @@
...
+        {/* Save and Cancel Buttons */}
+        <div className="flex justify-center">
+          <button onClick={handleSave} className="bg-green-500 hover:bg-green-600 text-white font-bold py-2 px-3 rounded-l-md">
+            ✓
+          </button>
+          <button onClick={onCancel} className="bg-yellow-500 hover:bg-yellow-600 text-white font-bold py-2 px-3 rounded-r-md">
+            ✕
+          </button>
+        </div>
+      </div>
```
norswap, line 62: At normal size, this overflows on my Macbook (1728 x 1117)

Screenshot 2024-02-16 at 17 05 05

There should be no scrolling in this panel, the input field should resize to fit the width of the panel, and you should use flexbox so that the buttons flow underneath the input box when space is limited.

packages/webapp/src/components/collection/deckPanel.tsx
```diff
@@ -0,0 +1,84 @@
...
+      <div className="flex justify-between items-center">
+        {/* Deck Name Input */}
+        <input
+          type="text"
+          value={deckName}
+          onChange={handleDeckNameChange}
+          style={{ outline: isDeckNameValid ? "none" : "2px solid red" }}
+          className="flex-grow px-2 py-2 border rounded-md text-gray-100 focus:outline-none focus:ring-2 focus:ring-blue-500 focus:border-transparent m-1.5"
+          placeholder=" Deck name"
+        />
```
norswap, line 51: The input font color here is a very light gray, it's impossible to read. Just a black or very dark gray will do. The placeholder font is a darker grey, that's fine (but if you use grey for the text, the difference has to be noticeable ... honestly just do black, black is fine here).

packages/webapp/src/components/collection/deckList.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+const DeckCollectionDisplay: React.FC<DeckCollectionDisplayProps> = ({ decks, onDeckSelect }) => {
+  return (
+      <div className="w-full flex flex-col items-center p-3">
+        {/* New Deck Button */}
+        <Link 
+          href={"/collection?newDeck=true"} 
+          className="w-full px-4 py-2 mb-2 border rounded-md text-gray-100 bg-purple-900 hover:bg-gray-500 font-bold text-center focus:outline-none focus:ring-2 focus:ring-blue-500 focus:border-transparent">
+          New Deck →
+        </Link>
```
norswap, line 18: 1. Make this more obviously clickable, maybe a button? This looks like a heading right now.
    I think maybe this should be button-ish, but the listed decks should not, they could be a list of underlined text follow [by one of those little arrows](https://gist.github.com/miguelmota/322c89234d60de578f37d3c6d30f7e41) (or from [react-icons](https://react-icons.github.io/react-icons/)  which we're already importing).
    
2. Make sure there is spacing under it. (1) will probably take care of that, but currently there's no spacing between "New Deck" and the list of decks.

packages/webapp/src/components/collection/deckPanel.tsx
```diff
@@ -0,0 +1,84 @@
...
+          </button>
+        </div>
+      </div>
+
+      {/* List of Cards in the Deck */}
+      <div className="mt-4 w-full">
+        {selectedCards.map((card, index) => (
+          <div 
+            key={index} 
+            className="p-2 border-b last:border-b-0 cursor-pointer hover:bg-gray-100"
```
norswap, line 69: I think we can lose the border-bottom here, it will look better.

packages/webapp/src/components/collection/deckPanel.tsx
```diff
@@ -0,0 +1,84 @@
...
+          </button>
+          <button onClick={onCancel} className="bg-yellow-500 hover:bg-yellow-600 text-white font-bold py-2 px-3 rounded-r-md">
+            ✕
+          </button>
+        </div>
+      </div>
+
+      {/* List of Cards in the Deck */}
+      <div className="mt-4 w-full">
+        {selectedCards.map((card, index) => (
```
norswap, line 66: Could you add an explanation text here for when this is empty? Just "Click on cards to add them to the deck." in a way that looks nice (color a bit muted but readable, good spacing around it).

packages/webapp/src/components/collection/deckPanel.tsx
norswap: I think we should also clarify this. Maybe move the buttons to the next line, replace ✓ with "Save deck" and ✕ with "Cancel".

Down the line, we'll gray out "save deck" if nothing changed, and pop a confirmation model if trying to cancel after making a change. But let's keep that for a further PR.

norswap (changes requested):
This is progressing nicely! We're dealing with the finer details now.

Also don't forget to rebase regularly!

norswap (changes requested):
The history is a bit of a mess, but I think everything from the first "make check issues" to the end can be squashed together via an interactive rebase. Let's sync on telegram, maybe do a call and I'll walk you through how to do that.

Regarding the UI, we're progressing but not quite there yet. The textfield still overflows and the buttons don't flow vertically when needed. I suggest playing with resizing the window to see how it behaves!

packages/webapp/src/components/collection/deckPanel.tsx
eviterin: The buttons are now on their own row and made explicit with "Save" and "Cancel".
Borders between cards are now removed.
Resizing looks good (to me) at least until it becomes very small.

image
image

Is button auto scaling supported right now? It seems that the font currently dictates the size of them, even with w-auto (see `components/ui/button`).

packages/webapp/src/components/collection/cardCollectionDisplay.tsx
```diff
@@ -0,0 +1,55 @@
...
+                  selectedCards.some(c => c.id === card.id) ? 'shadow-highlight shadow-orange-300' : '' // Highlight if selected
+                } hover:bg-slate-800 rounded-lg p-4 border-4 border-slate-900 grow w-[220px] max-w-[330px]`}
+                onMouseEnter={() => setSelectedCard(card)}
+                onClick={() => {
+                  if (isEditing) {
+                    onCardToggle(card) // Only toggle card selection when in editing mode
+                  }
+                }}
+              >
+                <Image className="aspect-square" src="/card_art/0.jpg" alt={card.lore.name} width={256} height={256} />
```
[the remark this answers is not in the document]
eviterin: I think I've now done the logic correctly. But `card-list.ts` currently has random card art and names.

norswap:
Global feedback:

- behaviour much better now!
- kinda looks ugly
    - not enough vertical spacing between text field & button
    - would be good to set a default deck name so we don't start with the red border
         - like Deck #X where X is the number of existing decks + 1
    - buttons should probably look more like the starting page button
         - in particular, reduce border size
         - would be good to have a different color for accept and cancel? (not sure)
     - add space between tick mark / cross mark and text
     - add vertical spacing between "New Deck" and listed decks
          - same, reduce border width
          - make "New Deck" and listed deck more visually distinct
               - keep "New Deck" as button but make listed deck more visually similar to the cards in the deck card list
- bugs
     - when editing an existing deck, the border shows as red until edited (but deck can still be edited & saved correctly)
- enhancements
      - grey out "save" if nothing in the deck changed, or if something is invalid (rn only invalid deck name)
      - the two panels (deck list & card lists) are inconsistent: deck list shows a scrollbar when it gets too small, whereas the card list doesn't
           - we should probably do the scrollbar for both (but keep the buttons reflowing when reducing size though!)
           - NOTE: chrome doesn't actually display the scrollbar but the content is scrollable horizontally in decklist, not in the card list still
                 - would be good to show the scroll bar on all browsers, but this is the least important item on this list
                 
Still, this is pretty usable, so I'm going to merge this now, so we can enable prettier globally. Then I think the next priorities should be:

- implement contract interop
- redo the layout according to #93 (we don't need the filters atm and it looks jank) — tackle the above items at the same item if still relevant in the new layout

norswap (approved):

Issue #32, opened 2023-06-03
Spin up own chain infrastructure


norswap:
For the tech demo, the game has to run somewhere. The long-term plan is to make it a rollup (settlement layer TBD), but in the short term, we want an easier solution.

For cost considerations, the easiest thing to do would be to spin our own Proof-of-Authoriy (PoA) chain. This would also enable to very easily make the game totally free [via a burner wallet](https://github.com/norswap/0xFable/issues/31).

Issue #31, opened 2023-06-03
Use a burner wallet


norswap:
Currently, every action in the game requires signing a transaction in the wallet, this is of course not acceptable.

There are multiple solutions to that issue, notably involving some form of account abstraction.

But the simplest solution by far is to set up a burner wallet: a private key derived locally (maybe from a single wallet signature) and kept in the browser storage, used to sign actions as players do them.

There is, to the best of my knowledge, no burner wallet library, however it's clear that plenty of games use them: Dark Forest, Conquest.eth, ... MUD comes bundled with one. So there are plenty of space to get inspiration from.

(It could be interesting to bundle whatever we end up implementing as an open-source library.)

Regarding practicality, this will need to be coupled with a faucet implementation for the local devnet. This can be fairly trivial: just bundle the standard devnet private keys and have them send some funds to the generated burner wallet.

For the eventual testnet, it's more difficult, and ties in to the question of how we will host the testnet. My preference would be set up a simple PoA chain, since that seems like the cheapest & simplest option for a no-stakes demo.

I think we could optimistically assume would not spam such a devnet. In any case, the spamming should not be too costly, it simply would make the game unusable (as the amounts dropped by the faucet would not be sufficient to play it).

The faucet implementation needs to be a bit more involved on such a devnet as well: a call to a separately hosted server holding the keys to an account with a huge gas token balance.

- [ ] Implement local devnet burner wallet
- [ ] Implement hosted testnet faucet

norswap:
We can use [these viem features](https://viem.sh/docs/accounts/privateKey.html#generating-private-keys) to create random private key / accounts.

For wagmi compatibility, we need to get the `publicClient` returned by `configureChain` ([docs](https://wagmi.sh/react/config)) and wrap it into a new function taking `{ chainId?: number }` as a parameter and either returning the result of `publicClient` (if we want to use the usual wallet connection stuff) or a new viem [PublicClient](https://viem.sh/docs/clients/public.html) if we want to use a burner wallet.

norswap:
An interesting system to look at: https://github.com/AudiusProject/hedgehog

norswap:
Let's use the account abstraction support in rollop to get this going. We should use [Permissionless.js simple-account support](https://docs.pimlico.io/permissionless/how-to/accounts/use-simple-account) (since simple-account that's what we use, along with a burner wallet). It should integrate neatly into wagmi/viem.

[...]

PR #110, opened 2024-01-21
Account abstraction

eerkaijun:

https://github.com/0xFableOrg/0xFable/issues/31 and https://github.com/0xFableOrg/0xFable/issues/32 (Issue 31 is dependent on issue 32 as account abstraction toolings cannot work on Anvil node)

## Describe Your Changes

1. Use rollop to be the dev node rather than Anvil. Include Makefile commands to start rollop node and account abstraction toolings.
2. Integrate permissionlessjs to deal with smart contract account
3. Users can opt to use a burner account which generates a private key and store on browser. The private key is then used as the signer of the smart contract account. Transactions are sponsored by the paymaster. (So the notion of "accounts" are abstracted away from end users)

## Checklist

## Testing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,6 @@ packages/circuits/out
 packages/circuits/trusted_setup
 packages/circuits/zkeys
 packages/webapp/public/proofs
+
+# Roll-op files
+roll-op/
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,17 @@ anvil:
 	cd packages/contracts && make anvil
 .PHONY: anvil
 
+# Runs rollop
+rollop:
+	@if [ ! -d "roll-op" ]; then \
+        git clone https://github.com/0xFableOrg/roll-op.git; \
+    fi
+	cd roll-op && ./rollop setup && ./rollop devnet
+
+# Runs account abstraction bundler and paymaster
+account-abstraction:
+	cd roll-op && ./rollop aa
+
 # Runs the webapp in dev mode.
 webdev:
 	cd packages/webapp && make dev
--- a/Makefile
+++ b/Makefile
@@ -21,14 +21,14 @@ anvil:
 
 # Runs rollop
 rollop:
-	@if [ ! -d "roll-op" ]; then \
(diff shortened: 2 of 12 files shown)


eerkaijun:
Just opening this as a draft PR - there are still a bunch of things to be done. Will be easier to collaborate and might need some help too. 

Currently main thing to figure out is how to connect `smartAccountClient` (defined [here](https://github.com/0xFableOrg/0xFable/blob/account-abstraction/packages/webapp/src/utils/accounts.ts)) to wagmi config. I think the idea is make it into a connector, though there are some typing incompatibility issues. 

Once that is figured out and tested, a bunch of other things to be done are:
- Clean up all use of Anvil
- Remove hardcoded values of localhost port (probably better to manage them through environment variables)
- Private key management and storage on the browser
- Some other clean up

norswap:
In the `BurnerConnector`, the wallet client is set here: https://github.com/0xFableOrg/0xFable/blob/master/packages/webapp/src/wagmi/BurnerConnector.ts#L63-L67

So you could assign the client there for that connector.

Connectors are set here: https://github.com/0xFableOrg/0xFable/blob/account-abstraction/packages/webapp/src/chain.ts#L78-L79

`getDefaultConnectors` in ConnectKit: https://github.com/family/connectkit/blob/35aeecd0cf12ef265beba724ae9b01fdc34b19b6/packages/connectkit/src/defaultConnectors.ts, populates them with a bunch of connectors defined in wagmi itself.

Here's the one used for Metamask: https://github.com/wevm/wagmi/blob/1.x/packages/connectors/src/injected.ts#L167-L181

The one from Coinbase Wallet is identical, but does not inherit `InjectedConnector`: https://github.com/wevm/wagmi/blob/1.x/packages/connectors/src/injected.ts#L167-L181 (the other ones, WalletConnect, Safe, ... seem similar, though maybe there's some slight subtlety I missed).

I think the move might be to monkey patch those to replace their `getWalletClient` function to return the original when not using account abstraction, and the AA one otherwise.

Issue #107, opened 2024-01-19
Automate card draw


norswap:
We should automatically submit the card draw transaction when it's our turn.

norswap:
fixed by #116

PR #116, opened 2024-02-01
Automate turn based card draw.

ultraviolet10:
Refers to #107.

Adds in a `useEffect` in `play.tsx` that is triggered whenever the turn switches, and triggers the `drawCard` transaction.
Removes the `Draw` button since we're now automating the action. 

To test: keep ending turns in each tab, the tx is triggered in the other tab immediately (logs confirm this).

@norswap, would be nice to use [this](https://ui.shadcn.com/docs/components/sonner) to indicate which card has been drawn? (when we have a definitive card list, with IDs and all).


norswap (changes requested):
I see one big problem in testing: the hand doesn't update after the draw. This could be a regression that was introduced earlier? It does update after a hard browser refresh.

I figured out what the issue was, but what do you think if I'd let you try to figure it out? I feel like it would be a good way to get a good grasp on the store logic / data architecture and the way these things update. Let me know what you think and if you want to search, don't hesitate to reach out as soon as you're stuck (or if you just want to use me as a duck to check your reasoning).

packages/webapp/src/store/derive.ts
```diff
@@ -128,9 +127,8 @@ export function getOpponentData(gdata: FetchedGameData|null, player: Address|nul
 /**
  * @see module:store/read#getPrivateInfo
  */
-export function getPrivateInfo(gameID: bigint|null, player: Address|null): PrivateInfo|null {
+export function getPrivateInfo(gameID: bigint|null, player: Address|null, privateInfoStore: PrivateInfoStore): PrivateInfo|null {
```
norswap, line 130: You will also need to change `getPrivateInfo` in `read.ts` to accommodate this. Let's get into the habit of running `make check` regularly. Will add an issue for adding github checks.

ultraviolet10: Yep have this as a change, haven't committed as working on the animation change suggested as well.

norswap:
TODO:
- change `read.ts` to match `derive.ts`
- run `make check`

After that, let's just merge this, I'll create an issue for the better UX of showing the status of drawing.

packages/webapp/src/store/derive.ts
norswap: Fantastic, keep up the good work :)

packages/webapp/src/pages/play.tsx
```diff
@@ -105,14 +106,27 @@ const Play: FablePage = ({ isHydrated }) => {
...
+
+  useEffect(() => {
+    // Automatically submit the card draw transaction when it's our turn
+    if (gameData && currentPlayer(gameData) === playerAddress && !cantDrawCard) {
+      toast.promise(doDrawCard, {
+        loading: "Your Turn - Drawing Card...",
+        success: () => {
+          return "Card Drawn Successfully!"
+        },
+        error: "Error Drawing Card",
```
norswap, line 126: We need to handle this better. The `drawCard` function will already call `setError` (in `write.ts`) to trigger the display of a global modal displaying the error. I think it's probably good to keep this for now. Can we just hide the toast in case of error?

Also if an error occurs, there's a question of what to do about card draw. The current implem will be stuck because the dependencies of the `useEffect` are not expected to change. But if the error is due to a temporary RPC outage, this is a problem.

We'll need to redisplay the draw button in this case. I'd do this by checking the return value of `drawCard` inside `doDrawCard` and if negative, set a state variable that controls the display of the old draw button.

packages/webapp/src/hooks/useScrollBox.ts
```diff
@@ -58,25 +71,50 @@ function useScrollBox(scrollRef: any) {
...
+  // Handles the wheel event to adjust the scrollLeft property, enabling horizontal scrolling.
   const handleScroll = (e: WheelEvent) => {
     if (scrollRef.current) {
       // Adjust the scrollLeft property based on the deltaY value
       scrollRef.current.scrollLeft += e.deltaY
     }
   }
 
+  // Responds to window resize events to update arrow visibility states.
   const handleResize = () => {
```
norswap, line 91: I really like the new comments in this file!

packages/webapp/src/hooks/useScrollBox.ts
```diff
@@ -58,25 +71,50 @@ function useScrollBox(scrollRef: any) {
...
     setShowLeftArrow(true)
     setShowRightArrow(true)
   }
 
+  // Smoothly scrolls to the rightmost end of the container and then back to the start.
```
norswap, line 96: Can you turn that to a `/** */` comment and add that it glows the last card?

packages/webapp/src/hooks/useScrollBox.ts
```diff
@@ -58,25 +71,50 @@ function useScrollBox(scrollRef: any) {
...
+  const smoothScrollToRightThenLeft = useCallback(() => {
+    const element = scrollRef.current
+    if (!element) return
+
+    const targetRight = element.scrollWidth - element.clientWidth
+    smoothScroll(targetRight, () => {
+      triggerLastCardGlow()
+      setTimeout(() => {
+        smoothScroll(0) // Scrolls back to the first element
```
norswap, line 105: Wait, why?

packages/webapp/src/hooks/useScrollBox.ts
```diff
@@ -2,27 +2,34 @@ import { useState, useEffect, useCallback } from "react"
...
   const checkArrowsVisibility = () => {
     if (!scrollRef.current) return
     const { scrollLeft, scrollWidth, clientWidth } = scrollRef.current
     setShowLeftArrow(scrollLeft > 0)
     setShowRightArrow(scrollLeft < scrollWidth - clientWidth)
   }
 
-  const smoothScroll = (target: number) => {
+  // Performs a smooth scrolling animation to a specified target position.
+  // Accepts a target scroll position and an optional callback to execute after completion.
```
norswap, line 31: Change to a `/** */` style comment.

packages/webapp/src/hooks/useScrollBox.ts
```diff
@@ -2,27 +2,34 @@ import { useState, useEffect, useCallback } from "react"
...
+  const smoothScroll = (target: number, callback?: () => void) => {
```
norswap, line 32: Shouldn't this be a `useCallback`?

packages/webapp/package.json
norswap, line 1: You need to update the pnpm lockfile when you update `package.json` otherwise `make setup` will complain that the lockfile is not up to date (since it does not specify the version for the new packages).

norswap (changes requested):
We're getting there :)

But sadly, the scroll animation did not work for me initially, neither Firefox nor Chrome.
(The card was drawn successfully and was displayed in the drawer though.)

Then somehow after a while it started working... No card glow though!

After further testing: there seems to be a dependency on being scrolled enough to the right. When I'm already at the right end of the card drawer, it works, or when I'm close to it. When I'm at the left end it never works.

Also, please remove the scroll back left haha, not sure why do that. Maybe we could scroll back to the previous location though? Feel like staying on the new card would be fine.

packages/webapp/src/hooks/useScrollBox.ts
ultraviolet10: I just thought it was a nice touch from an experience pov, i'll remove this :)

packages/webapp/src/actions/drawCard.ts
```diff
@@ -92,8 +92,6 @@ async function drawCardImpl(args: DrawCardArgs): Promise<boolean> {
   const cards = getCards()!
   console.log(`drew card ${cards[selectedCard]}`)
 
-  args.setLoading("Generating draw proof ...")
-
```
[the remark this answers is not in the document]
ultraviolet10: Something as simple as: 

```
function setLoading(message: string|null|undefined) {
  // Intentionally left blank to ignore any loading message.
  // This function accepts a message parameter but does nothing with it.
}
```
?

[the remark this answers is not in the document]
norswap: I think we should handle the case where it is set to null to make the toast go away.

ultraviolet10:
@norswap tried adding the timeout to the `refreshGameData` calls, didn't retrigger the useEffect that performs the `drawCard` functionality. ✌️

packages/webapp/src/actions/drawCard.ts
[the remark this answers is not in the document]
ultraviolet10: The toast disappears in a couple of seconds as is.

pnpm-lock.yaml
norswap, line 1: The lock still doesn't seem okay, try running `make setup` to confirm. Mine still says "Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with packages/webapp/package.json"

You can run `make install` to fix, though it's not ideal (as it might updates other dependencies, I think?), but let's do it now.
Normally, just using `pnpm install NEW_PKG` in the dir of a package (like `packages/webapp`) should be okay and update the lockfile (which needs to be committed).

packages/webapp/src/pages/_app.tsx
norswap, line 1: There seems to be a regression, probably introduced by the move to shadn where if the global error modal triggers, it does not display.

The browser console gives me

```
setting error modal: {"title":"Contract execution error","message":"Transaction reverted (drawInitialHand) with InvalidProof(undefined)","buttons":[{"text":"Dismiss"}]}
```

but the error modal itself is not displayed (or maybe it is but is hidden behind the normal game-creation modal, though a quick look at the DOM makes it seem like this isn't the case).

(As for the error itself, I'm investigating.)

We can leave this for a further issue (but to be prioritized!) and merge this without the fix since it isn't really related to what we're doing here.

packages/webapp/src/actions/drawCard.ts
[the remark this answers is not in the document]
norswap: I'm going to make the call we need to dismiss it — as if it overlaps with the card being highlighted, which is really "eh".

Also, we should "pop up" the hand (as if it was hovered by the cursor, even if it isn't) at least for the time of the card highlight. I would make the card highlight time a tad longer as well (maybe 3s instead of 2.5?)

norswap (changes requested):
Alright:

- fix package-lock
- don't forget to rebase!
- kill toast upon draw + pop up card drawer
- make check

and we should be good to go!

norswap (approved):

norswap:
Great work, merging this (after rebasing on top of master) 🔥

I think we could improve it further.

e.g. the card appears before the animation in a way that is noticeable, would be great if the animation was more instantaneous with that event, and would also be great if the toast was dismissed immediately when that happens.

I think the main issue there is that the actual update occurs from the `refreshGameState` whereas animation etc trigger from getting receipts from the transaction, which (at least locally) occurs after. I think this might need to happen as part of a bigger update on how we can signal those things via the first thing that occurs (state update OR getting receipts).

Also testing this I noticed the issue here #128

PR #125, opened 2024-02-21
Add github CI checks for linting

### LINKED ISSUE #119: Add Github CI checks for linting etc
We should run the global makefile's `make check` in [github CI](https://docs.github.com/en/apps/creating-github-apps/writing-code-for-a-github-app/building-ci-checks-with-a-github-app) and ensure it exits with a zero error code (meaning everything passed as expected).

Here's [an example of the same thing in the roll-op repo](https://github.com/0xFableOrg/roll-op/blob/master/.github/workflows/build.yml).

Note that we don't need to "build": the contracts tests will build them automatically, and the typescript code doesn't need a build step. HOWEVER, we will need `make setup` first to install all dependencies.

ptisserand:

This PR add a github CI workflow which will run `make check`
Closes #119 
## Checklist

## Testing

I have tested using https://github.com/nektos/act

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -1,0 +1,21 @@
+name: Check project
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: pnpm/action-setup@v3
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: nightly
+      - run: make setup
+      - run: make check
+
--- a/packages/webapp/src/pages/_app.tsx
+++ b/packages/webapp/src/pages/_app.tsx
@@ -76,7 +76,7 @@ const ComponentWrapper = ({
   const errorConfig = useErrorConfig()
 
     // todo @eviterin: i've understood it so that decks are stored on chain. thus, below part is not going to be needed.
-    const testCards = [];
+    const _testCards = [];
     const [decks, setDecks] = useState<Deck[]>([]);
     //
 
--- a/packages/webapp/src/pages/editor.tsx
+++ b/packages/webapp/src/pages/editor.tsx
@@ -90,7 +90,7 @@ const Editor: FablePage = ({ decks, setDecks, isHydrated }) => {
     } else {
       setOriginalDeckIndex(null) // Reset if not editing an existing deck
     }
-  }, [router.query.index, decks])
+  }, [decks])
 
   const isCardInDeck = (cardToCheck: Card) => {
     return deck.some(cardInDeck => cardInDeck.id === cardToCheck.id)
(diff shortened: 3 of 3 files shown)


norswap:
Fantastic, let's merge it!

norswap:
Ayoo, it fails: https://github.com/0xFableOrg/0xFable/actions/runs/8000911253

> Type error: Cannot find module 'contracts/out/deployment.json' or its corresponding type declarations.

This makes perfect sense. We either need to commit a dummy deployment file to the repo, or run Anvil and deploy in the action itself.

Actually, I suggest adding a command to the github action that simply creates an empty JSON (file containing `{}`) at  `contracts/out/deployment.json`. That seems like the easiest way out of this.

We're also getting this (unrelated) warning:

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

We need to update the foundry-toolchain action [to the latest version](https://github.com/foundry-rs/foundry-toolchain/releases).

Finally, I suggest testing the action in a github fork, as clearly local testing is not perfect ;)

norswap:
I forced-pushed the merge away, so as not to bother the other PRs with failed CI runs.

ptisserand:
> > Type error: Cannot find module 'contracts/out/deployment.json' or its corresponding type declarations.
> 
> This makes perfect sense. We either need to commit a dummy deployment file to the repo, or run Anvil and deploy in the action itself.
> 
> Actually, I suggest adding a command to the github action that simply creates an empty JSON (file containing `{}`) at `contracts/out/deployment.json`. That seems like the easiest way out of this.
> 

I have fixed this issue with [adding the following step](https://github.com/ptisserand/0xFable/commit/bc34cf377f01608e501bd50c0fe1021ca265fdfb):
```
      - name: Fake contract deploymment
        run: mkdir -p packages/contracts/out && echo "{}" > packages/contracts/out/deployment.json
```

> We're also getting this (unrelated) warning:
> 
> > Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
> 
> We need to update the foundry-toolchain action [to the latest version](https://github.com/foundry-rs/foundry-toolchain/releases).
> 
We are already using the latest `foundry-toolchain` action, I have updated `actions/checkout` to fix this warning:
https://github.com/ptisserand/0xFable/actions/runs/8013673319

> Finally, I suggest testing the action in a github fork, as clearly local testing is not perfect ;)
Sure, sorry for this mistake.

@norswap Should I create a new PR ?

norswap:
> We are already using the latest foundry-toolchain action

Are you sure? This specifies v1 but latest version is v1.1.1, and v1.1 moves version to Node 16 to Node 20.
Not sure if these versions are wildcards.

> @norswap Should I create a new PR ?

Yes please :)

ptisserand:
> > We are already using the latest foundry-toolchain action
> 
> Are you sure? This specifies v1 but latest version is v1.1.1, and v1.1 moves version to Node 16 to Node 20. Not sure if these versions are wildcards.
> 
`v1` is a "moving tag", it's the same changeset as `v1.1.1`

> > @norswap Should I create a new PR ?
> 
> Yes please :)

Here is link to the new PR: https://github.com/0xFableOrg/0xFable/pull/127

Issue #35, opened 2023-06-03
Detect offline client


norswap:
Display some info if we detect we can't get online.

The simplest thing is to use [`navigator.online`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/onLine) however this is only `false` if the LAN is unreachable. Meaning the internet might be down but local networking up and this would still be `true`.

Another option to improve on this is to detect a RPC query that fails all its allotted retries.

Blazing-Mike:
I want to take this issue, but running this project is a hassle. i have followed the read me strictly but the `make anvil` command not running getting an error. is there any other guide to running this project.

norswap:
There isn't, but instead of complaining, a more productive approach would be to report to error so we can try to diagnose it.

Since all `make anvil` does is run `anvil` it either means `make` or `anvil` are not installed correctly. You can try running `make --version` or `anvil --version` to confirm.

I don't mean to be rude, but I think getting stuck at this stage maybe indicates you might not be a good fit for this project!

Blazing-Mike:
Thank you, i do not take offense either, i am just genuinely trying to contribute to this project as part of the ODHack and probably get experience with web3 projects. 

I finally got `make anvil` running but when i run `make webapp` i have and error that says `Module not found: Can't resolve 'src/generated'`. I check the src folder and saw that there's no src/generated file. so i am confused what command during the setup creates this file. 

Cc: @norswap

norswap:
All good! Feel free to work on this issue!

knownothing10:
I’d like to work on this.

ryzen-xp:
Can I take care of this issue?

VyuduInc:
Can I be assigned to this?

sir-etiosa:
I'd like to take this issue.

PrathamJha1:
Would love to contribute here , given my experience as a JavaScript and Typescript developer working in the Web3 space for over a year.

chiscookeke11:
I’d like to help with this.

Richardudeh4:
Hi,

I’ll use navigator.onLine to check basic connectivity and implement a retry mechanism with Axios or Fetch API for failed RPC queries. A modal will be displayed on the home and play pages when connectivity is lost. The modal will clearly inform users of any internet issues.

wengDavo:
Could I try solving this?

abhishek-nexgen-dev:
Mind if I try this one?

PR #129, opened 2024-02-26
detect offline client

Blazing-Mike:

Link related issues!
i ran into git conflict so i had to pull. open another PR
#126 #35 

## Checklist

## Testing

I implemented a modal to display when our app is offline. so how did i test this, i tested this by throttling the network in my devtools and also disconneCTING my device from wifi and this modal appears on the index.tsx and play.tsx page. when i toggle back or reconnect, the modal disaappears

CC: @norswap

### DIFF
--- a/packages/webapp/src/hooks/useOfflineCheck.ts
+++ b/packages/webapp/src/hooks/useOfflineCheck.ts
@@ -1,0 +1,33 @@
+import { useEffect, useState } from "react"
+
+import { toast } from "sonner"
+
+function useOfflineCheck() {
+    const [isOnline, setIsOnline] = useState(true)
+
+    useEffect(() => {
+        const handleOnline = () => setIsOnline(true)
+        const handleOffline = () => setIsOnline(false)
+
+        window.addEventListener("online", handleOnline)
+        window.addEventListener("offline", handleOffline)
+
+        return () => {
+            window.removeEventListener("online", handleOnline)
+            window.removeEventListener("offline", handleOffline)
+        }
+    }, [])
+
+    useEffect(() => {
+        if (!isOnline) {
+            toast.error("App is offline. Please check your internet connection.", {
+                dismissible: false,
+                duration: Infinity,
+            })
+        } else {
+            toast.dismiss()
+        }
+    }, [isOnline])
+}
+
+export default useOfflineCheck
--- a/packages/webapp/src/pages/_app.tsx
+++ b/packages/webapp/src/pages/_app.tsx
@@ -15,6 +15,7 @@ import jotaiDebug from "src/components/lib/jotaiDebug"
 import { GlobalErrorModal } from "src/components/modals/globalErrorModal"
 import { Toaster } from "src/components/ui/sonner"
 import { useIsHydrated } from "src/hooks/useIsHydrated"
+import useOfflineCheck from "src/hooks/useOfflineCheck"
 import { useErrorConfig } from "src/store/hooks"
 
 import "src/styles/globals.css"
@@ -33,6 +34,7 @@ export type FablePage = NextPage<{ isHydrated: boolean }>
 // =================================================================================================
 
 const MyApp: AppType = ({ Component, pageProps }) => {
+    useOfflineCheck()
     return (
         <>
             <Head>


packages/webapp/src/pages/index.tsx
```diff
@@ -37,6 +38,8 @@ const Home: FablePage = ({ isHydrated }) => {
   const isRightNetwork = !notConnected && chainSupported
   const isWrongNetwork = !notConnected && !chainSupported
 
+    useOfflineCheck();
+
```
norswap, line 42: Sorry I forgot this one is the last review, could you add this to `_app.tsx` instead, so it applies to all pages?

packages/webapp/src/hooks/useOfflineCheck.ts
```diff
@@ -0,0 +1,73 @@
...
+  const endpoint = "https://www.google.com";
+  try {
+    const response = await fetch(endpoint, {
+      method: "HEAD",
+      cache: "no-cache",
+      mode: "no-cors",
+    });
+    return response.ok;
+  } catch (error) {
+    throw error; // Rethrow for retries
```
norswap, line 14: This is redundant, you could remove the whole `try-catch` statement and it would work the same.

packages/webapp/src/hooks/useOfflineCheck.ts
```diff
@@ -0,0 +1,73 @@
...
+        setTimeout(() => {
+          performNetworkCheck()
+            .then(() => handleOnline()) // online
+            .catch(() => {
+              setRetries(retries + 1); // Retry if error occurs
+            });
+        }, retryDelay);
+      } else {
+        setIsOffline(true);
+        toast.error("App is offline. Please check your internet connection.");
```
norswap, line 46: Add an option object here with `{ dismissible: false, duration: Infinity }` which prevents the toast from closing.

norswap: I'm wondering if we need this at all — can't we just handle the "online" and "offline" events? On "online" do what you do in `handleOnline`, on "offline" do `setIsOffline` and the toast. What's the reasoning here? Is there some documentation around the behaviour and reliability of these events?

packages/webapp/src/hooks/useOfflineCheck.ts
```diff
@@ -0,0 +1,73 @@
...
+function useOfflineCheck(options: any = {}) {
+  const {
+    // Optional configuration properties:
+    threshold = 1000, // Minimum interval between offline checks (ms)
+    maxRetries = 3, // Maximum number of retries before assuming offline
+    retryDelay = 1500, // Delay between retries (ms)
+  } = options;
+
+  const [isOffline, setIsOffline] = useState(false);
+  const [retries, setRetries] = useState(0);
```
norswap, line 27: These two being React state is not needed I think, as it means the `useOfflineCheck` we will rerender each time we increment the amount of retries or set `isOffline` — this is unnecessary, the `toast.error` function will trigger a rerender itself and nothing else is needed. You can make those `let` variables and mutate them directly.

packages/webapp/src/hooks/useOfflineCheck.ts
```diff
@@ -0,0 +1,73 @@
...
+    // Initial network check:
+    performNetworkCheck()
+      .then(() => {
+        setIsOffline(false);
+        setRetries(0);
+      })
+      .catch(() => {
+        setIsOffline(true);
+        toast.error("App is offline. Please check your internet connection.");
+      });
```
norswap, line 62: The toast is triggered many times, which we should fix.

I'm wondering if this isn't the problem, as this is rerun every time `useOfflineCheck` is rerun, which is every time that the page is rendered.

Another possible source of the issue is that the "offline" event is fired multiple times? Is that the case?

packages/webapp/src/hooks/useOfflineCheck.ts
```diff
@@ -0,0 +1,73 @@
...
+    return response.ok;
+  } catch (error) {
+    throw error; // Rethrow for retries
+  }
+}
+
+function useOfflineCheck(options: any = {}) {
+  const {
+    // Optional configuration properties:
+    threshold = 1000, // Minimum interval between offline checks (ms)
```
norswap, line 21: This is never used.

packages/webapp/src/hooks/useOfflineCheck.ts
```diff
@@ -0,0 +1,73 @@
...
+      setRetries(0); // Reset retries on reconnection
+    };
+
+    const handleOffline = () => {
+      if (retries < maxRetries) {
+        setTimeout(() => {
+          performNetworkCheck()
+            .then(() => handleOnline()) // online
+            .catch(() => {
+              setRetries(retries + 1); // Retry if error occurs
```
norswap, line 41: I'm quite dubious on the retry logic here, as we don't actually enqueue any retry! The only way retries can be attempted is if the "offline" event fire multiple times. Is that the intention here? But if so, the `retryDelay` variable is a little bit misleading, as it simply adds to the delay between multiple firings of "offline".

norswap (changes requested):

packages/webapp/src/pages/index.tsx
Blazing-Mike: sure

packages/webapp/src/hooks/useOfflineCheck.ts
Blazing-Mike: i think i figured why the toast is rendering multiple times, it is being called both in the `performNetworkCheck()` function and the handleOffline.

packages/webapp/src/hooks/useOfflineCheck.ts
Blazing-Mike: i think i was overthinking this one and i have brought forward logic i was using before you introduced the sonner component. we don't need the isOffline state again.

packages/webapp/src/hooks/useOfflineCheck.ts
Blazing-Mike: for the retry logic, we only want to show our app is offline if `retries === maxRetries. so the reason for that.

Blazing-Mike:
@norswap please i need a review on this PR.

norswap:
Hey, time is a bit tight right now, but I'll get to it soon. Don't worry, this'll be considered for ODHack.

Blazing-Mike:
I understand. Thank you.

packages/webapp/src/hooks/useOfflineCheck.ts
```diff
@@ -0,0 +1,60 @@
...
+            }) // online
+            .catch(() => {
+              const newRetry = retries + 1;
+              setRetries(newRetry); // Retry if error occurs
+            });
+        }, retryDelay);
+      } 
+    };
+
+    window.addEventListener("online", handleOffline);
```
norswap, line 41: It seems unecessary to recheck the network state when receiving "online".

packages/webapp/src/hooks/useOfflineCheck.ts
```diff
@@ -0,0 +1,60 @@
...
+              setRetries(0);
+              toast.dismiss();
+            }) // online
+            .catch(() => {
+              const newRetry = retries + 1;
+              setRetries(newRetry); // Retry if error occurs
+            });
+        }, retryDelay);
+      } 
+    };
```
norswap, line 39: The retry logic still doesn't make sense.

1. `handleOffline` is called
2. `performNetworkCheck` is called
2. in case of failure (`catch`) the retry count is incremented by 1
3. ??? — nothing in the code makes `performNetworkCheck` or `handleOffline` be called again

norswap: Do we really need retry logic in the first place? We could just assume offline after a single failure.

The risk of assuming after single failure is that it's a fluke and the `offline` event and subsequent `online` events are never sent, in which case we stay in offline mode forever.

Do we even need `performNetworkCheck` at all? Can't we only rely on `online` and `offline`?

It's possible that these are not always reliable and it's recommend to perform explicit checks, but I'd like to hear your reasoning about this + some documentation showing that indeed `online` / `offline` is not to be relied on.

packages/webapp/src/hooks/useOfflineCheck.ts
norswap: Did you try the codebase and shut down your wifi to see if this works correctly?

norswap (changes requested):

packages/webapp/src/hooks/useOfflineCheck.ts
Blazing-Mike: this is checked to dismiss the toast.

packages/webapp/src/hooks/useOfflineCheck.ts
Blazing-Mike: Hmmmm, it might be complicating things. The first option I opted for was using the `online` and `offline `, but seeing some documentation that suggests online/offline might not be relied on that's why I opted for doing a network check and retrying the request if it fails the first time.

In my opinion, the unreliability of the online/offline event might be rare and we can tackle it with listening for both online and offline. I implemented a version that uses the online and offline and they both work. I will push the changes for review. 

Cc: @norswap

norswap:
Hey, any updates on this?

Blazing-Mike:
Yess. I have fixed the implementation to just rely on offline and online events. And it works when I switched off WiFi and turned it on.

Blazing-Mike:
cc @norswap

norswap:
Sorry, I missed that extra commit. Looks good, could you rebase on top of master and make sure `make check` passes? (You'll want to run `make format` first.)

Blazing-Mike:
Hi norswap.Just reminding you of this PR, make check have passed.

norswap:
Hey, I see you merged instead of rebasing, do you think you could get that fixed? If not, it's okay, just let me know and I'll perform the rebase myself :)

Blazing-Mike:
running into some issues with it(. could you help please. thank you.

Blazing-Mike:
it's me again 🙂...just wanna check if the or is merged. Have a productive week ahead. 

Cc: @norswap

norswap:
Hey, I'm unfortunately super buy and hadn't time to deal with this, but I'll get it sorted!

norswap (approved):
Let's go! 🚀

Blazing-Mike:
finally 😃!!I feel so elated looking forward to contributing more to this project.

PR #130, opened 2024-02-27
Add global formatting for the webapp.

ultraviolet10:
Adds a global format command that runs prettier's `--write` on all specified file types. 

Uses the following options within the config (`prettierrc.json`):

- `"semi": false`: prevents addition of semicolons on save
- `"trailingComma": "es5"`: https://prettier.io/docs/en/options.html#trailing-commas
- `"singleQuote": false`: uses double quotes for imports/elsewhere.
- `"tabWidth": 2`: @norswap did you want this as 4? 
- `"useTabs": true`: https://prettier.io/docs/en/options.html#tabs
- `"printWidth": 100`: increase the number of characters that prettier will try to wrap around, prevents the dependency arrays of `useEffect` and `useCallback` hooks from splitting into a column unless absolutely needed (https://prettier.io/docs/en/options.html#print-width). 

Also added `eslint-config-prettier` to be used as a plugin as the recommended mediator between linting and formatting.


ultraviolet10:
there's another interesting plugin we could use: https://www.npmjs.com/package/eslint-plugin-simple-import-sort

this helps sort imports in different files according to your specs, here's an example one: 

```
"simple-import-sort/imports": [
      "error",
      {
        "groups": [
          // Packages. `react` related packages come first.
          [
            "^react",
            "^next",
            "@metamask|@walletconnect|clsx|@reduxjs|async-mutex|flowbite-react|redux-persist|pubnub|redux",
            "^(ethers|google|lodash|bn.js|styled-components|child_process|cluster|console|constants|crypto|dgram|dns|domain|events|fs|http|https|module|net|os|path|punycode|querystring|readline|repl|stream|string_decoder|sys|timers|tls|tty|url|util|vm|zlib|freelist|v8|process|async_hooks|http2|perf_hooks)"
          ],
          // Internal packages.
          ["^@?\\w"],
          // Parent imports. Put `..` last.
          ["^\\.\\.(?!/?$)", "^\\.\\./?$"],
          // Other relative imports. Put same-folder imports and `.` last.
          ["^\\./(?=.*/)(?!/?$)", "^\\.(?!/?$)", "^\\./?$"],
          // Style imports.
          ["^.+\\.s?css$"],
          // Side effect imports.
          ["^\\u0000"]
        ]
      }
      ```

**Comment by ultraviolet10** [comment]
As well as: https://github.com/tailwindlabs/prettier-plugin-tailwindcss

**Comment by norswap** [comment]
> `"tabWidth": 2`: @norswap did you want this as 4?

Yes let's go for 4!

> "useTabs": true: https://prettier.io/docs/en/options.html#tabs

This is always religious, but I do prefer spaces myself. I guess with prettier it matters less, but it ensures proper display on Github for instance.

Any counter here? Like would you strongly prefer a 2 or 8 spaces local display?

File: .github/workflows/check.yaml [hunk 1]
```diff
@@ -21,3 +22,7 @@ jobs:
       - run: make setup
       - run: make check
 
+      - name: Install dependencies for webapp
+        run: cd packages/webapp && pnpm install
```
norswap, line 26: this should be covered by `make setup` above I believe

.github/workflows/check.yaml
```diff
@@ -21,3 +22,7 @@ jobs:
...
+      - name: Check formatting with Prettier
+        run: cd packages/webapp && npx prettier --check "**/*.{js,jsx,ts,tsx,json,css}"
```
norswap, line 28: This should be folded into `make check` in the webapp Makefile (which will cause it be checked by the global Makefile `make check` and thus covered by the above)

Makefile
```diff
@@ -40,6 +40,12 @@ check:
 	cd packages/webapp && make check
 .PHONY: check
 
+# Performs code formatting for the webapp files and contracts in their respective directories.
+format:
+	cd packages/webapp && make format
+	cd packages/contract && make check
```
norswap, line 46:
```suggestion
	cd packages/contract && make format
```

packages/webapp/Makefile
```diff
@@ -26,4 +26,9 @@ lint:
 
 # Runs code quality checks.
 check: lint build
-.PHONY: check
\ No newline at end of file
+.PHONY: check
+
+# Runs prettier formatting across webapp files with specified file extensions.
+format: 
+	npx prettier --write "**/*.{js,jsx,ts,tsx,json,css}"
```
norswap, line 33:
```suggestion
	pnpm prettier --write "**/*.{js,jsx,ts,tsx,json,css}"
```

norswap (changes requested):

ultraviolet10:
> > `"tabWidth": 2`: @norswap did you want this as 4?
> 
> Yes let's go for 4!
> 
> > "useTabs": true: https://prettier.io/docs/en/options.html#tabs
> 
> This is always religious, but I do prefer spaces myself. I guess with prettier it matters less, but it ensures proper display on Github for instance.
> 
> Any counter here? Like would you strongly prefer a 2 or 8 spaces local display?

No hard preference here, 4 is good!

packages/webapp/.prettierrc.json
norswap, line 1: isn't 150 too much? I like 100 or 120

packages/webapp/Makefile
```diff
@@ -25,5 +25,11 @@ lint:
 .PHONY: lint
 
 # Runs code quality checks.
-check: lint build
-.PHONY: check
\ No newline at end of file
+check: 
+	pnpm eslint . && pnpm prettier --check "**/*.{js,jsx,ts,tsx,json,css}"
+.PHONY: check
```
norswap, line 30: you can put both of these on different lines (no need for &&)

norswap (approved):
Approving, but address comments!

PR #132, opened 2024-03-04
Add import and style sorting config + capabilities.

ultraviolet10:
Adds [auto class sorting with Prettier](https://tailwindcss.com/blog/automatic-class-sorting-with-prettier) for tailwind styles and [import sorting](https://github.com/lydell/eslint-plugin-simple-import-sort) with ESLint.

Also adds the eslint `fix` command to the `format` command in the webapp Makefile, since `check` will now show an import sort errors.


packages/webapp/.eslintrc.cjs
```diff
@@ -3,9 +3,11 @@ const config = {
 	extends: ["next/core-web-vitals", "plugin:@typescript-eslint/recommended", "prettier"],
 	parser: "@typescript-eslint/parser",
 	parserOptions: {
-		project: "./tsconfig.json",
-	},
-	plugins: ["@typescript-eslint"],
+    project: "./tsconfig.json",
+		"sourceType": "module",
+    "ecmaVersion": "latest",
```
norswap, line 8: What's with the weird indentation here?

packages/webapp/.eslintrc.cjs
```diff
@@ -26,6 +28,31 @@ const config = {
 				caughtErrorsIgnorePattern: "^_",
 			},
 		],
+		"import/first": "error",
+    "import/newline-after-import": "error",
+    "import/no-duplicates": "error",
+		"simple-import-sort/imports": [
+			"error",
+			{
```
norswap, line 36: Weird alignment, is this not caught by the formatter?

packages/webapp/package.json
```diff
@@ -52,8 +51,11 @@
         "autoprefixer": "^10.4.16",
         "eslint": "^8.52.0",
         "eslint-config-next": "^13.5.6",
+        "eslint-config-prettier": "^9.1.0",
+        "eslint-plugin-import": "^2.29.1",
+        "eslint-plugin-simple-import-sort": "^12.0.0",
         "postcss": "^8.4.31",
-        "prettier": "2.8.8",
+        "prettier-plugin-tailwindcss": "^0.5.11",
```
norswap, line 58: Why are we removing prettier here?

packages/webapp/.eslintrc.cjs
```diff
@@ -26,6 +28,31 @@ const config = {
...
+					// Internal packages.
+					["^@?\\w"],
+					// Parent imports. Put `..` last.
+					["^\\.\\.(?!/?$)", "^\\.\\./?$"],
+					// Other relative imports. Put same-folder imports and `.` last.
+					["^\\./(?=.*/)(?!/?$)", "^\\.(?!/?$)", "^\\./?$"],
+					// Style imports.
+					["^.+\\.s?css$"],
+					// Side effect imports.
+					["^\\u0000"]
```
norswap, line 52: One rule that would be great is to make sure the internal imports, which we prefix with `src/` are in their own separate section (after the "internal packages"  would be ideal — we should also change that comment to actually say "external packages" (stuff like wagmi etc))

norswap (changes requested):

packages/webapp/.eslintrc.cjs
ultraviolet10: It doesn't seem to appear on my local repo. 😶

packages/webapp/package.json
ultraviolet10: added this back, accidentally committed something else I was testing.

packages/webapp/.eslintrc.cjs
norswap: It's because of mixed tabs and spaces. We should add `.cjs` and `.mjs` files to the list of those who get formatted by prettier, cand you do that?

packages/webapp/src/components/cards/cardContainer.tsx
```diff
@@ -1,12 +1,15 @@
...
-import BoardCard from "./boardCard"
-import HandCard from "./handCard"
 import { CSS } from "@dnd-kit/utilities"
+
+import { CardPlacement } from "src/store/types"
 import { convertStringToSafeNumber } from "src/utils/js-utils"
 
+import BoardCard from "./boardCard"
+import DraggedCard from "./draggedCard"
+import HandCard from "./handCard"
```
norswap, line 11: These should use the `src/` format. I know it's not formatting related, but could you fix it?

I think we could use [this type of rule](https://stackoverflow.com/questions/71118709/eslint-only-allow-relative-import-paths-not-absolute) to prevent them, but I wonder if we'd need to be able to make exception to import files form other packages (e.g. deployment addresses) — I think we could get around that by copying files however. Could you look into it?

packages/webapp/src/components/hand.tsx
```diff
@@ -1,11 +1,15 @@
...
+
+import { CancellationHandler } from "src/components/modals/loadingModal"
 import { CardPlacement } from "src/store/types"
-import CardContainer from "./cards/cardContainer"
 import { convertBigIntArrayToStringArray } from "src/utils/js-utils"
-import { CancellationHandler } from "src/components/modals/loadingModal"
+
+import useScrollBox from "../hooks/useScrollBox"
+
+import CardContainer from "./cards/cardContainer"
```
norswap, line 12: Some more that need fixing.

packages/webapp/src/components/modals/globalErrorModal.tsx
```diff
@@ -1,8 +1,9 @@
+import { useEffect, useState } from "react"
+
+import { Button } from "src/components/ui/button"
 import { ErrorConfig } from "src/store/types"
 
 import { Dialog, DialogContent, DialogTitle } from "../ui/dialog"
```
norswap, line 6: Anotha one.

packages/webapp/src/components/modals/inGameMenuModalContent.tsx
```diff
@@ -1,8 +1,9 @@
 import Link from "next/link"
 
-import { DialogDescription, DialogTitle } from "../ui/dialog"
 import { Button } from "src/components/ui/button"
 
+import { DialogDescription, DialogTitle } from "../ui/dialog"
```
norswap, line 5: Anotha one.

packages/webapp/src/components/modals/joinGameModal.tsx
```diff
@@ -1,22 +1,23 @@
...
+import { useCancellationHandler } from "src/hooks/useCancellationHandler"
+import * as store from "src/store/hooks"
 import { GameStatus } from "src/store/types"
+import { setError } from "src/store/write"
+import { isStringPositiveInteger, parseBigIntOrNull } from "src/utils/js-utils"
 import { navigate } from "src/utils/navigate"
-import { useCancellationHandler } from "src/hooks/useCancellationHandler"
-import { concede } from "src/actions/concede"
 
 import { Dialog, DialogContent, DialogDescription, DialogTitle, DialogTrigger } from "../ui/dialog"
```
norswap, line 20: Moar

norswap (changes requested):
Added comments to fix some import issues unrelated to sorting + fix the indentation problem by adding cjs/mjs to prettier.

packages/webapp/src/components/modals/globalErrorModal.tsx
ultraviolet10: isn't this the expected format?

packages/webapp/src/components/cards/cardContainer.tsx
ultraviolet10: added this rule, running `make check` and `make format` doesn't pop up any errors.

packages/webapp/.eslintrc.cjs
```diff
@@ -17,6 +17,13 @@ const config = {
         "@typescript-eslint/no-empty-function": "off",
         "@typescript-eslint/no-explicit-any": "off",
         "@typescript-eslint/ban-ts-comment": "off",
+        "no-restricted-imports": "off",
+        "@typescript-eslint/no-restricted-imports": [
+            "error",
+            {
+                patterns: ["!./*", "!../*"],
+            },
+        ],
```
norswap, line 26: So I was confused why this would work because this is the pattern used in the link I gave to ONLY ALLOW relative links.

Turns out:

- ` make check`  does not actually run the eslint checks, it needs to be `pnpm next lint --max-warnings 0`  and not ` pnpm eslint` (but let's actually call `make lint` from `make check`, that way we avoid duplication)

- Even so, this should forbid every single one of our imports... turns out it does not work, probably `!` is not a thing as an operator.

- BUT, if you remove the `!`, then it does exactly what we want, and indeed catches a few relative imports that slipped through the cracks :)

Can you fix `make check`, this file and eliminate the remaining relative imports? And also add a comment in this file to explain we forbid relative imports and absolute imports starting with `src/` should be used instead?

Oh, also you'll need to exclude `generated.ts` from eslint (`ignorePatterns` option).

I see we're also excluding `src/hooks/useScrollBox.ts` from linting, seems to be about React `useEffect` dependencies, would be good to remove it from the ignore list and either fix or simply add comments to disable eslint on those `useEffect` statements.

norswap (changes requested):
The import pattern rule uncovered some issues

norswap (approved):
I think this is good to go?

ultraviolet10:
> I think this is good to go?

tumblr_lyzwwppm5P1qh28hmo1_400

Issue #117, opened 2024-02-14
Implement player handles


norswap:
Two key features:

- resolve players ENS if they have one, use https://viem.sh/docs/ens/actions/getEnsName.html

- Add a `PlayerHandle` contract where users can register a single handle for their address (they should be able to change it as well)
     - Handles should be unique
     - Add a `checkHandleValidity` function that restricts names (e.g. only ASCII, minimum 5 characters, maybe not contain dots (to avoid ENS / other name system clashes)
     - Add a `checkPlayerEligibility` that can be always `true` for now but down the line will only enable players that have completed some milestones (e.g. minimum amount of matches, minimum spend, ...) to register handles to avoid squatting.
     - If a user has an ENS, let them select whether they want to use their handle or their ENS.
     - Add an admin role that can be used later to avoid impersonation and squatting of popular names.

The frontend must be modified to display player handles (or ENS).

faytey:
can i hop on this?

norswap:
Go ham :)

faytey:
Thanks I'd work on this, for better context, I'd be writing a cairo contract to handle Player Ens/profile and updating the frontend showing their ens right?

norswap:
No, we're using Solidity on this game as we target an EVM rollup chain. Cairo contracts only work on Starknet.

saimeunt:
@faytey have you started working on this?
I'd like to take this one if you change your mind and work on something else please.

faytey:
Just seeing this notification @saimeunt yes you can go ahead. @norswap I’d take up other issues thanks

faytey:
If this issue is still up, i would love to continue working on it

saimeunt:
Hey I'm no longer actively working on this feel free to pick it @faytey

faytey:
> Hey I'm no longer actively working on this feel free to pick it @faytey

Cool, @norswap is this okay?

norswap:
Yes, go for it!

faytey:
Hi Sir @norswap I made a PR to this issue here [117](https://github.com/0xFableOrg/0xFable/issues/117) I need your review and comments sir.

faytey:
@norswap 

> Hi Sir @norswap I made a PR to this issue here [117](https://github.com/0xFableOrg/0xFable/issues/117) I need your review and comments sir.

norswap:
Sorry, things have been hectic these days. I will try to review this as soon as possible, but it might still take a few days!

PR #136, opened 2024-05-20
Adding Player Handles

faytey:

Link related issues!
Issue [117](https://github.com/0xFableOrg/0xFable/issues/117)

## Describe Your Changes
I added the smart contract to implement Player handles and wrote the tests to check the functions work properly

## Checklist

## Testing

### DIFF
--- a/packages/contracts/src/PlayerHandle.sol
+++ b/packages/contracts/src/PlayerHandle.sol
@@ -1,0 +1,106 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
+import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
+
+interface IENSResolver {
+    function name(bytes32 node) external view returns (string memory);
+}
+
+error InvalidHandle();
+error HandleAlreadyTaken();
+error PlayerNotEligible();
+
+contract PlayerHandle is Ownable {
+    using Strings for uint256;
+
+    IENSResolver public ensResolver;
+    mapping(address => string) private handles;
+    mapping(string => address) private handleOwners;
+    mapping(address => bool) private useEns;
+
+    event HandleRegistered(address indexed player, string handle);
+    event HandleChanged(address indexed player, string newHandle);
+    event UseEnsSet(address indexed player, bool useEns);
+
+    constructor(address _ensResolver) {
+        ensResolver = IENSResolver(_ensResolver);
+    }
+
+    function registerHandle(string memory handle) external {
+        if (!checkHandleValidity(handle)) {
+            revert InvalidHandle();
+        }
+        if (handleOwners[handle] != address(0)) {
+            revert HandleAlreadyTaken();
+        }
+        if (!checkPlayerEligibility(msg.sender)) {
+            revert PlayerNotEligible();
+        }
+
+        if (bytes(handles[msg.sender]).length > 0) {
+            handleOwners[handles[msg.sender]] = address(0);
+        }
+
+        handles[msg.sender] = handle;
+        handleOwners[handle] = msg.sender;
+
+        emit HandleRegistered(msg.sender, handle);
+    }
+
+    function changeHandle(string memory newHandle) external {
+        if (!checkHandleValidity(newHandle)) {
+            revert InvalidHandle();
+        }
+
+        if (handleOwners[newHandle] != address(0)) {
+            revert HandleAlreadyTaken();
+        }
+
+        string memory oldHandle = handles[msg.sender];
+        handles[msg.sender] = newHandle;
+        handleOwners[newHandle] = msg.sender;
+        handleOwners[oldHandle] = address(0);
+
+        emit HandleChanged(msg.sender, newHandle);
+    }
+
+    function setUseEns(bool _useEns) external {
+        useEns[msg.sender] = _useEns;
+        emit UseEnsSet(msg.sender, _useEns);
+    }
+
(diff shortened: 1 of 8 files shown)


faytey:
I need your review on this before proceeding, I'm encountering errors setting up the web

faytey:
> I need your review on this before proceeding @norswap , Also I'm encountering errors setting up the webapp

Hello @norswap still awaiting your review

package-lock.json
norswap, line 1: Could you remove this file? We use `pnpm` for package management and we already have a pnpm-lock file in place!

packages/contracts/src/PlayerHandle.sol
```diff
@@ -0,0 +1,84 @@
...
+
+        handles[msg.sender] = handle;
+        handleOwners[handle] = msg.sender;
+
+        emit HandleRegistered(msg.sender, handle);
+    }
+
+    function changeHandle(string memory newHandle) external {
+        require(checkHandleValidity(newHandle), "Invalid handle");
+        require(handleOwners[newHandle] == address(0), "Handle already taken");
```
norswap, line 44: Could you use custom errors for those?

packages/contracts/src/PlayerHandle.sol
```diff
@@ -0,0 +1,84 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {Ownable} from "openzeppelin/access/Ownable.sol";
+import {Strings} from "openzeppelin/utils/Strings.sol";
+
+interface IENSResolver {
+    function getEnsName(address addr) external view returns (string memory);
```
norswap, line 8: That doesn't seem to be the actual ENS resolver interface, which uses `name()` for lookups, see here: https://docs.ens.domains/resolvers/public

We also probably won't deploy on Ethereum, which is the only chain that has this resolver at the moment. ENS has announced that L2 support is coming but it's not there yet. I think it's fine to keep the resolver interface for now, at worst we can stub it with a permissioned registry that we sign on — let's just make the interface the same as the real thing.

packages/contracts/src/PlayerHandle.sol
```diff
@@ -0,0 +1,84 @@
...
+        }
+        return handles[player];
+    }
+
+    function checkHandleValidity(string memory handle) public pure returns (bool) {
+        bytes memory b = bytes(handle);
+        if (b.length < 5) return false;
+        for (uint256 i; i < b.length; i++) {
+            bytes1 char = b[i];
+            if (char < 0x20 || char > 0x7E || char == 0x2E) {
```
norswap, line 74: Note to myself: will probably want to change this to only support alphanumerics & dashes, underscores, square brackets (for cute clan names), spaces (but not more than one in a row, and not at the end or at the start of the handle), as well as put a max character limit.

Fine to leave this as is for this PR.

packages/contracts/src/test/PlayerHandle.t.sol
```diff
@@ -0,0 +1,78 @@
...
+    mapping(address => string) private ensNames;
+
+    function setEnsName(address addr, string memory name) external {
+        ensNames[addr] = name;
+    }
+
+    function getEnsName(address addr) external view override returns (string memory) {
+        return ensNames[addr];
+    }
+}
```
norswap, line 17: Could you spin this out in a separate file (in the test repo), and add a deployment of this + `PlayerHandle` to the `Deploy.s.sol`? You'll also want to add an entry for `PlayerHandle` in `deployment.ts` on the frontend.

norswap (changes requested):
Great work! Sorry again for the delay. Could you make the requested changes? I'll make sure to prioritize reviewing them now.

faytey:
Hi Sir I'd get started on these and update, sorry for the delay

custosDiretriz:
Hi Sir, I've made the necessary corrections and pushed

faytey:
Hi Sir, I have made requested corrections

norswap:
Also it seems you've made changes to the lockfile, could just revert it to the master version so that no changes show? The current file seems incorrect in some way, as shown by the failing CI check.

faytey:
> Also it seems you've made changes to the lockfile, could just revert it to the master version so that no changes show? The current file seems incorrect in some way, as shown by the failing CI check.

Hi Sir, I didn't make any changes to the lockfile and can't seem to understand how to rectify it.

norswap:
That was probably accidental, you can see it is changed in "Files changes" here on Github.

Fixing it requires a little bit of git foo:
You can revert it by using `git checkout master -- pnpm-lock.yaml` then `git add pnpm-lock.yaml`.
You could then git commit and push that, but let's squash all your commit together:
`git reset --soft HEAD~7 && git commit`

(The 7 here is the number of commits you've made, so if you add additional commits in between you need to update that number.)

Then finally `git push --force-with-lease` to replace the commits here with the single squashed commit.

faytey:
> That was probably accidental, you can see it is changed in "Files changes" here on Github.
> 
> Fixing it requires a little bit of git foo: You can revert it by using `git checkout master -- pnpm-lock.yaml` then `git add pnpm-lock.yaml`. You could then git commit and push that, but let's squash all your commit together: `git reset --soft HEAD~7 && git commit`
> 
> (The 7 here is the number of commits you've made, so if you add additional commits in between you need to update that number.)
> 
> Then finally `git push --force-with-lease` to replace the commits here with the single squashed commit.

done sir

faytey:
Hello sir @norswap

norswap:
Thanks! Did some extra cleanup and merged here: https://github.com/0xFableOrg/0xFable/pull/137

PR #147, opened 2026-04-10
fix: only show drop indicator when card play is legal

### LINKED ISSUE #121: Add indicators for illegal card play
Currently the game allows playing a single card every turn.

Two things are thus illegal:
- playing a card when it's not your turn
- playing a card when you've already played one

But the interface currently shows the green drop background in all of these scenarios. This should be changed so that this drop background only appears in situation where playing a card is legal. The condition is probably already in the code somewhere, since we do not trigger a `playCard` transaction when it is not possible, which just need to extend the condition to the display of the green background.

Ai-chan-0411:
Closes #121

## Problem
The green drop background for card placement was showing in all drag scenarios, even when playing a card is illegal (not the player's turn, or already played a card this turn).

## Solution
Added conditions to the `playerActive` check in `playerBoard.tsx`:

1. **Is the local player's board** — drop indicator only shows on your own board
2. **Is the local player's turn** — uses `useCurrentPlayerAddress()` to check whose turn it is
3. **Game step is PLAY** — checks `gameData.currentStep === GameStep.PLAY` to ensure the player hasn't already played a card this turn

The existing `playCard` action already enforces these conditions (line 59 of `playCard.ts`), but the UI was not reflecting them visually.

## Changes
- `packages/webapp/src/components/playerBoard.tsx`: Import `GameStep`, add `useGameData()` and `useCurrentPlayerAddress()` hooks, update `playerActive` condition

/attempt 121


Ai-chan-0411:
Closing due to no maintainer review in 48h+. Thank you for considering!

